PR #4, opened 2024-09-17
platform detection

josibake:


TheCharlatan:
Closing, found a better way to detect if the compiler is clang and subsequently link lib++.

PR #6, opened 2025-05-08
Build script improvements

josibake:
While trying to build this in NixOS, I ran into a few snags in the build script. Here are my fixes; tried to make sure they were overall improvements as opposed to hacks, but feel free to suggest a better approach if you have one in mind.


TheCharlatan:
Thanks Josie, will take a look once I'm home again. Can you re-ttigger the CI now by pushing again?

libbitcoinkernel-sys/build.rs
```diff
@@ -52,12 +52,19 @@ fn main() {
         .status()
         .unwrap();
 
-    // Add the lib directory to the search path
-    let lib_dir = install_dir.join("lib");
+    // Check multiple possible lib directories
+    let possible_lib_dirs = vec![install_dir.join("lib"), install_dir.join("lib64")];
```
TheCharlatan, line 56: Mmh, it is weird that you run into this at all. `cmake` by default puts everything into /lib on install, and we are installing into a local directory here. I don't think we should be making an exception here. Does the build pass in your environment if you instead add
```
        .arg("-DCMAKE_INSTALL_LIBDIR=lib")
```
To the cmake configuration above? That should override any interfering environment variables.

josibake: Nice, that works! Also agree, this is much cleaner. I've update the first commit to use this approach.

josibake:
@TheCharlatan did a bunch of reading and digging to see how I could be smarter about choosing the c++ stdlib, but there doesn't seem to be a clean solution to this. What I have now still feels hacky, but it seems better than assuming "libc++" when compiling with clang (which is not the case when compiling with clang on linux).

I also prefer this to what I had before, which was using a custom env var to override it. From what I can tell, 99% of the time a user on macos will be using clang++ and libc++, and in all other cases (clang on linux), its most likely going to be stdlibc++.

Thoughts?

TheCharlatan:
> @TheCharlatan did a bunch of reading and digging to see how I could be smarter about choosing the c++ stdlib, but there doesn't seem to be a clean solution to this. What I have now still feels hacky, but it seems better than assuming "libc++" when compiling with clang (which is not the case when compiling with clang on linux).

I've been doing so reading what other wrapper crates are doing here, and it is a pretty mixed bag. Some use the `cc` crate that has some "automagic" stuff built in, some just roll everything themselves, including passing in all the necessary build flags, and some do a mix, like I am attempting here. I guess that is the price you pay for typically having as little build configuration as possible in Rust projects. The current code is already hacky enough, with just invoking some other random compiler to figure out what might be installed on the system. If this helps solve your problem for now though, I'd be happy to merge this.

I just pushed a commit to master for a macos CI job. Can you rebase on that?

josibake:
Rebased on master! 

> If this helps solve your problem for now though, I'd be happy to merge this.

This solves the problem I'm having for now, and would likely avoid any surprises in the future for anyone else compiling on linux but using clang as the compiler. So a bit hacky, but doesn't seem to be a regression over what is here now. I'll open an issue for improving the build script and keep it as a low priority thing to hack on as I get ideas. Leveraging the `cc` crate more seems promising.

josibake:
Pending CI, I'd say this is good to merge and I've opened https://github.com/TheCharlatan/rust-bitcoinkernel/issues/7 to track and discuss further (low priority, though, so no rush).

PR #12, opened 2025-08-11
Refactor: Adapt to btckApi changes and implement Rust idioms

alexanderwiederin:
# Adapt to btckApi changes and implement Rust idioms

## Breaking Changes
- **New spent outputs hierarchy**: Replaced `BlockUndo` with a cleaner `BlockSpentOutputs` -> `TransactionSpentOutputs` -> `Coin` design
- **Rust naming conventions**: Removed `get_` prefixes from all methods (`get_value()` -> `value()`, `get_hash()` -> `hash()`, etc.)
- **Zero-copy borrowing**: Most methods now return `RefType<T, Owner>` for borrowed access

## New Features
- **RefType for memory-safe borrowing**: Introduced `RefType<'a, T, L>` for zero-copy access to borrowed data with compile-time lifetime safety
- **Transaction/block iteration**: Added `output_count()`, `transaction_count()`, `output()`, `transaction()` methods
- **Clone support**: All data types now implement `Clone` using the underlying C copy functions
- **Dual conversion APIs**: Both explicit methods (`to_bytes()`) and conversion traits (`From<T> for Vec<u8>`)
- Updated and improved documentation

## Memory Safety & Performance
- **Zero-copy by default**: Methods like `script_pubkey()`, `output_at()`, `coin()` return borrowed references
- **Explicit copying**: Call `to_owned()` on RefType when you need owned data that can outlive the parent
- **Compile-time safety**: Borrow checker prevents use-after-free with borrowed references

## Internal Improvements
- Updated C API bindings from `kernel_` to `btck_` prefixes
- Added input validation to `verify` function
- Smart ownership tracking leverages C library's `m_owned` flags

## Migration
```rust
// Before
let value = output.get_value();
let hash = block.get_hash();

// After  
let value = output.value();
let hash = block.hash();

// RefType usage patterns
let script_ref = output.script_pubkey();    // RefType<ScriptPubkey, TxOut>
let script_bytes = script_ref.to_bytes();   // Direct method call via Deref
let owned_script = script_ref.to_owned();   // Convert to owned when needed

### CODE REVIEW

File: src/lib.rs [hunk 1]
```diff
@@ -594,53 +604,109 @@ impl TryFrom<&[u8]> for ScriptPubkey {
...
+    /// Returns the amount of this output in satoshis.
+    pub fn value(&self) -> i64 {
+        unsafe { btck_transaction_output_get_amount(self.inner) }
     }
 
-    /// Get the script pubkey of this output
-    pub fn get_script_pubkey(&self) -> ScriptPubkey {
+    /// Returns the script pubkey that defines how this output can be spent.
+    ///
+    /// The returned script pubkey is independent of this output's lifetime.
```
TheCharlatan, line 656: I don't think this is true, the returned script pubkey here should still be dependent on the output's lifetime. Or am I missing something? I think this similarly needs to either return a ref type, manage the lifetime explicitly, or add phantom data.

TheCharlatan (approved):
lgtm

PR #19, opened 2025-09-08
refactor: adapt to btck_api changes kernelApi_63 and introduce trait-based API

alexanderwiederin:
## Refactor: adapt to btck_api changes kernelApi_63

Adapts to C API changes and introduces a trait-based API.

**API Changes**
* Replace generic RefType<T,L> with specific reference types (ScriptPubkeyRef, TxOutRef, TransactionRef, etc.)
* Add trait-based API with *Ext traits (ScriptPubkeyExt, TxOutExt, TransactionExt, etc.) for operations across owned/borrowed types
* Update BlockTreeEntry::prev() to return Option instead of Result
* Simplify API by removing unnecessary unwrap() calls and Result returns
* Update BlockTreeEntry to use lifetimes tied to ChainstateManager instead of owned pointers

**Code Organization**
* Add prelude module for trait imports

**Breaking Changes**
* RefType<T,L> is removed - specific reference types should be used instead
* Some method signatures changed to accept trait objects instead of concrete types.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,6 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 
-use std::borrow::Borrow;
 use std::ffi::{c_char, c_void, CString, NulError};
 use std::marker::PhantomData;
 use std::{fmt, panic};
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -831,57 +832,89 @@ impl Drop for ScriptPubkey {
...
+
+impl<'a> Copy for ScriptPubkeyRef<'a> {}
+
+/// Common operations for transaction outputs, implemented by both owned and borrowed types.
+pub trait TxOutExt {
+    /// Returns a raw pointer to the underlying C object.
+    fn as_ptr(&self) -> *const btck_TransactionOutput;
+
+    /// Returns the amount of this output in satoshis.
+    fn value(&self) -> i64 {
```
rustaceanrob, line 907: Why is this signed out of curiosity?

TheCharlatan: It's what its type is in /consensus: https://github.com/bitcoin/bitcoin/blame/master/src/consensus/amount.h#L12

TheCharlatan:
lgtm

Tried to re-write some of the logic, but could not come up with something really satisfying. There is definitely a lot of repetition at the moment that would ideally be avoided. Some derive macros could help with that eventually. I'll still wait with publishing a new release until the other PR is in. It would also be nice to re-work our examples to make better use of the nested types.

PR #30, opened 2025-09-17
Chore(btck_api): update to btck_api 65

alexanderwiederin:
Updates to API changes from btck_api_65 release

## Changes
- Replace chain height deduction with `btck_chain_get_height` kernel API call
- Disable IPC support in libbitcoinkernel builds with -DENABLE_IPC=OFF CMake flag


TheCharlatan (approved):
ACK e19803ce36435786ed32abb3920a7badc4d12e4c

PR #34, opened 2025-09-22
Feat(core): Implement BlockHash wrapper for C API type (kernelApi_66)

alexanderwiederin:
## Implement BlockHash wrapper for C API type (kernelApi_66)

Implements wrapper for the `btck_BlockHash` opaque pointer type. Adapts to https://github.com/TheCharlatan/bitcoin/tree/kernelApi_66

**API Changes**
* Add `BlockHash` owned type wrapping `*mut btck_BlockHash`
* Add `InvalidLength { expected: usize, actual: usize }` variant to `KernelError`
* Update `Block::hash()` to return the new `BlockHash` instead of struct with `.hash` field
* Update `BlockTreeEntry::block_hash()` to return the new `BlockHash`
* Implement standard traits: `PartialEq`, `Eq`, `Debug`
* Provide conversions: `From<[u8; 32]>`, `TryFrom<&[u8]>`, `From<BlockHash> for [u8; 32]`

**Breaking Changes**
* `BlockHash` no longer has a public `.hash` field - use `.to_bytes()` method instead

### DIFF
--- a/src/core/block.rs
+++ b/src/core/block.rs
@@ -1,13 +1,14 @@
 use std::{ffi::c_void, marker::PhantomData};
 
 use libbitcoinkernel_sys::{
-    btck_Block, btck_BlockSpentOutputs, btck_Coin, btck_TransactionSpentOutputs, btck_block_copy,
-    btck_block_count_transactions, btck_block_create, btck_block_destroy, btck_block_get_hash,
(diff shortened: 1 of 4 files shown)


TheCharlatan (approved):
ACK 9b99b8f774e076d60158175245ac47981bd37951

PR #37, opened 2025-09-23
Feat(core): Complete Core Type Pattern and Add Trait Verification

alexanderwiederin:
## Feat(core): Complete Core Type Pattern and Add Trait Verification

This PR establishes a consistent Core wrapper pattern across all types and adds trait verification tests.

Related to https://github.com/TheCharlatan/rust-bitcoinkernel/issues/23

### Changes

**Implement Send + Sync for reference types**
- All reference types (`ScriptPubkeyRef`, `TransactionRef`, `CoinRef`, etc.) now implement `Send + Sync`
- Safe because refs wrap const pointers to immutable data protected by the Kernel's internal synchronization

**Add FromMutPtr to complete owned type pattern**
- Implemented `FromMutPtr` for `TransactionSpentOutputs`, `Coin`, `ScriptPubkey`, `Transaction`, and `TransactionOutput`
- All owned types now implement: `Clone`, `Drop`, `Send`, `Sync`, `AsPtr`, and `FromMutPtr`

**Add macro-based trait verification tests**
- New test macros `test_owned_trait_requirements!`, `test_ref_trait_requirements!`, `test_owned_clone_and_send!` and `test_ref_copy!`

### DIFF
--- a/src/core/test_utils.rs
+++ b/src/core/test_utils.rs
@@ -41,4 +41,69 @@ macro_rules! test_ref_trait_requirements {
 }
 
 #[cfg(test)]
-pub(crate) use {test_owned_trait_requirements, test_ref_trait_requirements};
+macro_rules! test_owned_clone_and_send {
+    ($test_name:ident, $obj1:expr, $obj2:expr) => {
+        #[test]
+        fn $test_name() {
+            let obj1 = $obj1;
+            let obj2 = $obj2;
+            let clone1 = obj1.clone();
+            let clone2 = obj2.clone();
+
+            assert_ne!(
+                obj1.as_ptr() as *const u8,
+                clone1.as_ptr() as *const u8,
+                "Clone should create independent object"
+            );
+            assert_ne!(
+                obj2.as_ptr() as *const u8,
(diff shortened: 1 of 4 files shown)


src/core/test_utils.rs
```diff
@@ -0,0 +1,47 @@
+#[cfg(test)]
+macro_rules! test_owned_trait_requirements {
+    ($test_name:ident, $owned:ty, $ffi_type:ty) => {
+        #[test]
+        fn $test_name() {
+            use crate::ffi::sealed::{AsPtr, FromMutPtr};
+
+            fn assert_clone<T: Clone>() {}
+            #[allow(drop_bounds)]
```
TheCharlatan, line 9: I think we can just skip the check for `drop`. If it is not implemented correctly, we leak memory, which hopefully either the sanitizers, or valgrind will pick up.

TheCharlatan (approved):
Nice, lgtm.

TheCharlatan (approved):
ACK 8bd8fae940942d316ecf70f9132276258280a2db

PR #44, opened 2025-10-08
Chore(state): Add unit tests for Chainstate Module

alexanderwiederin:
### Chore(state): Add unit tests for Chainstate module

Related to https://github.com/TheCharlatan/rust-bitcoinkernel/issues/23

### Changes

- ChainstateManagerOptions creation with valid and invalid paths
- ChainstateManager creation

### DIFF
--- a/src/state/chainstate.rs
+++ b/src/state/chainstate.rs
@@ -217,3 +217,60 @@ impl Drop for ChainstateManagerOptions {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{ChainType, ContextBuilder};
+    use tempdir::TempDir;
+
+    fn create_test_context() -> Context {
+        ContextBuilder::new()
+            .chain_type(ChainType::Regtest)
+            .build()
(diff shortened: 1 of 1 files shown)


src/state/chainstate.rs
```diff
@@ -216,3 +216,142 @@ impl Drop for ChainstateManagerOptions {
...
+    fn test_chainstate_manager_options_new() {
+        let context = create_test_context();
+        let (_temp_dir, data_dir, blocks_dir) = create_test_dirs();
+
+        let opts = ChainstateManagerOptions::new(&context, &data_dir, &blocks_dir);
+        assert!(opts.is_ok());
+    }
+
+    #[test]
+    fn test_chainstate_manager_options_set_worker_threads() {
```
TheCharlatan, line 258: I'm not sure if all these setter tests are useful, since they don't actually exercise anything in a hot path. This just ensures that the program doesn't crash, which in my view should just be a fuzz test. I would keep the basic new tests and those where we actually check errors. The rest seems a bit gratuitous.

PR #56, opened 2025-10-16
Kernel api 74

sedited:


TheCharlatan:
> Do we need as_ref for the BlockHash?

Yes, pushed.

alexanderwiederin:
Yap, sorry!

PR #63, opened 2025-10-23
Refactor(Log): Adapt Logger to new C API (subtree-76)

alexanderwiederin:
### Changes

Update logger to match new libbitcoinkernel C API where logging options are set via `btck_logging_set_options()` instead of being passed to `btck_logging_connection_create()`.

- Add `set_logging_options()` and `LoggingOptions` struct
- Add `Logger::new_with_options()` convenience method


TheCharlatan (approved):
lgtm

PR #64, opened 2025-10-27
Fix(core): add BlockHashExt to public exports

alexanderwiederin:
### Changes:

Exports BlockHashExt trait in core module and includes it in the prelude for easier access to block hash extension methods.


TheCharlatan (approved):
lgtm

PR #74, opened 2025-11-03
Feat(ci): Add fuzz corpus verification workflow

alexanderwiederin:
### Feat(ci): Add fuzz corpus verification workflow

Related to https://github.com/TheCharlatan/rust-bitcoinkernel/issues/35

### Changes

Add a new CI job that verifies fuzz corpus files from the [qa-assets repository](https://github.com/alexanderwiederin/qa-assets) against all fuzz targets. The job:

- Checks that corpus directories exist and contain files
- Verifies all corpus files exectue without crashes with -runs=0 (no new input generation)
- Uploads crash artifacts on failure for debugging

This ensures corpus files remain valid as the codebase evovles and helps catch regressions that might cause previously-valid inputs to crash.

*note: more fuzz tests will be added in later PRs in similar fashion.*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,57 +183,6 @@ jobs:
     - name: Build and test
       run: cargo test -vv
 
-  fuzzing:
-    name: Check Fuzzing Setup
-    runs-on: ubuntu-latest
-
-    steps:
-    - uses: actions/checkout@v3
-
-    - name: Install Rust toolchain
-      uses: actions-rs/toolchain@v1
-      with:
-        profile: minimal
-        toolchain: nightly
-        override: true
-        targets: x86_64-unknown-linux-gnu
-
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -233,3 +233,92 @@ jobs:
         cargo fuzz run block_roundtrip -- -max_total_time=20
         cargo fuzz run chainman_process_block -- -max_total_time=20
         cargo fuzz run script_verify -- -max_total_time=20
+
+  fuzz-corpus:
```
TheCharlatan: I think we can just completely replace the existing fuzzing job with this one. Can you also activate the address sanitizer for this job?

alexanderwiederin: Cool! Removed the fuzz setup test in the next commit and added the sanitizer for the new. Thanks for the feedback.

PR #75, opened 2025-11-03
Enable chainman fuzzer to run in parallel

sedited:
Make sure that every process gets its own data directory


brianstohl87-cpu (commented):
Yes sir

brianstohl87-cpu (commented):
It's better

PR #88, opened 2025-11-05
Chore(state): Improve Documentation for Chainstate module

alexanderwiederin:
Related to https://github.com/TheCharlatan/rust-bitcoinkernel/issues/27

### Changes

Adds module level documentation and improves struct and method docs for the chainstate.rs module.

Second attempt for https://github.com/TheCharlatan/rust-bitcoinkernel/pull/81

### DIFF
--- a/src/state/chainstate.rs
+++ b/src/state/chainstate.rs
@@ -385,17 +385,56 @@ impl Drop for ChainstateManager {
     }
 }
 
-/// Holds the configuration options for creating a new [`ChainstateManager`]
+/// Builder for configuring and creating a [`ChainstateManager`].
+///
+/// Provides a fluent interface for configuring how the chainstate manager
+/// will initialize and operate. Options control database locations, in-memory
+/// operation, worker thread allocation, and database initialization behavior.
+///
+/// # Usage
+/// Create a builder using [`ChainstateManager::builder`], configure options
+/// by chaining method calls, then call [`build`](Self::build) to create the
+/// chainstate manager.
+///
+/// # Example
+/// ```no_run
+/// use bitcoinkernel::{ChainType, ChainstateManager, ContextBuilder, KernelError};
+///
+/// # fn main() -> Result<(), KernelError> {
+/// let context = ContextBuilder::new()
+///     .chain_type(ChainType::Regtest)
+///     .build()?;
+///
+/// let chainman = ChainstateManager::builder(&context, "/data", "/blocks")?
+///     .worker_threads(4)
+///     .chainstate_db_in_memory(true)
+///     .block_tree_db_in_memory(true)
+///     .build()?;
+/// # Ok(())
+/// # }
+/// ```
 pub struct ChainstateManagerBuilder {
     inner: *mut btck_ChainstateManagerOptions,
 }
 
 impl ChainstateManagerBuilder {
-    /// Create a new option
+    /// Creates a new chainstate manager builder.
+    ///
+    /// This is typically called via [`ChainstateManager::builder`] rather than directly.
     ///
     /// # Arguments
-    /// * `context` -  The [`ChainstateManager`] for which these options are created has to use the same [`Context`].
-    /// * `data_dir` - The directory into which the [`ChainstateManager`] will write its data.
+    /// * `context` - The [`Context`] that configures chain parameters and
+    ///   notification callbacks. It is recommended to keep the context in scope
+    ///   for the lifetime of the [`ChainstateManager`].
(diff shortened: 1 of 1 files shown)


src/state/chainstate.rs
```diff
@@ -1,3 +1,27 @@
+//! Chainstate manager for Bitcoin Core validation and chain state management.
```
TheCharlatan: I would drop `Bitcoin Core` here. I don't think that makes it clearer.

src/state/chainstate.rs
```diff
@@ -1,3 +1,27 @@
...
+//!
+//! # Example
+//! ```no_run
+//! use bitcoinkernel::*;
+//!
+//! let context = ContextBuilder::new()
+//!     .chain_type(ChainType::Regtest)
+//!     .build()?;
+//!
+//! let opts = ChainstateManagerOptions::new(&context, "/data", "/blocks")?
```
TheCharlatan: Looking at this again, I think we should call this `ChainstateManagerBuilder` instead and implement a proper builder pattern on it.

src/state/chainstate.rs
```diff
@@ -74,14 +80,23 @@ impl ProcessBlockResult {
     }
 }
 
-/// The chainstate manager is the central object for doing validation tasks as
-/// well as retrieving data from the chain. Internally it is a complex data
-/// structure with diverse functionality.
+/// The chainstate manager handles Bitcoin block validation and chain state.
+///
+/// This is the primary interface for interacting with the Bitcoin blockchain,
```
TheCharlatan: s/blockchain/chain/

src/state/chainstate.rs
```diff
@@ -100,12 +128,39 @@ impl ChainstateManager {
...
+    /// Attempts to validate and add the block to the chain. The block goes
+    /// through full consensus validation including proof-of-work, transaction
+    /// validity, and script verification.
+    ///
+    /// # Arguments
+    /// * `block` - The [`Block`] to process
+    ///
+    /// # Returns
+    /// A [`ProcessBlockResult`] indicating whether the block was:
+    /// - Newly accepted and added to the chain ([`ProcessBlockResult::NewBlock`])
```
TheCharlatan: Confusingly, new block does not indicate whether the block was added to the chain. It is only used to indicate if it is committed to disk.

I think it would be good to mention that to good the full validity information, users are required to register with the validation interface's block checked callback.

src/state/chainstate.rs
```diff
@@ -46,10 +46,16 @@ use crate::{
...
+///
+/// Indicates whether a block was accepted, rejected, or was already known
+/// to the chainstate manager.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ProcessBlockResult {
-    /// Block was accepted and is new
+    /// Block passed validation and was added to the chain.
+    ///
+    /// This indicates the block is new and has been successfully validated
+    /// against all consensus rules.
```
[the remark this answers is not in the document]
alexanderwiederin: Kept it short for now with `Block was accepted and is new to the chainstate manager`

src/state/chainstate.rs
```diff
@@ -175,11 +323,38 @@ impl ChainstateManager {
         Ok(unsafe { BlockSpentOutputs::from_ptr(inner) })
     }
 
+    /// Get a reference to the currently active blockchain.
+    ///
+    /// Returns the active chain, which represents the chain with the most
+    /// accumulated proof-of-work. This is the canonical chain used for validation
+    /// and represents the current consensus state of the network.
```
TheCharlatan: I think this is too strong, maybe just drop this last line?

src/state/chainstate.rs
TheCharlatan: looks good.

src/state/chainstate.rs
```diff
@@ -251,7 +549,34 @@ impl ChainstateManagerBuilder {
...
+    /// and means the database will be lost when the process exits.
+    ///
+    /// # Arguments
+    /// * `chainstate_db_in_memory` - If true, run the chainstate database in memory
+    ///
+    /// # Memory Requirements
+    /// The chainstate database size varies by network:
+    /// - Mainnet: ~5-10 GB
+    /// - Testnet: ~1-2 GB
+    /// - Regtest: Typically very small
```
TheCharlatan: I think metrics that are guaranteed to change over time should not be in the docs.

TheCharlatan (approved):
lgtm

I likely missed something, but let's keep going.

PR #89, opened 2025-11-05
Implement builder pattern for ChainstateManagerOptions

sedited:
And add drop implementation for ContextBuilder.

### DIFF
--- a/src/state/chainstate.rs
+++ b/src/state/chainstate.rs
@@ -66,6 +66,22 @@ unsafe impl Send for ChainstateManager {}
 unsafe impl Sync for ChainstateManager {}
 
 impl ChainstateManager {
+    pub fn builder(
+        context: &Context,
+        data_dir: &str,
+        blocks_dir: &str,
+    ) -> Result<ChainstateManagerBuilder, KernelError> {
+        ChainstateManagerBuilder::new(context, data_dir, blocks_dir)
+    }
+
+    pub fn new(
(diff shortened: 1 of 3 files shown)


src/state/chainstate.rs
```diff
@@ -66,16 +66,6 @@ unsafe impl Send for ChainstateManager {}
 unsafe impl Sync for ChainstateManager {}
 
 impl ChainstateManager {
-    pub fn new(chainman_opts: ChainstateManagerOptions) -> Result<Self, KernelError> {
```
alexanderwiederin:
```rust 
impl ChainstateManager {
    pub fn builder(
        context: &Context, 
        data_dir: &str, 
        blocks_dir: &str
    ) -> Result<ChainstateManagerBuilder, KernelError> {
        ChainstateManagerBuilder::new(context, data_dir, blocks_dir)
    }
}
```
and we can do:
```rust 
ChainstateManager::builder(&context, &data_dir, &blocks_dir)?
    .worker_threads(4)
    .build()?
```

What do you think?

TheCharlatan: Yeah, might make sense to implement a default builder() and new() method for convenience.

alexanderwiederin (approved):
ACK https://github.com/TheCharlatan/rust-bitcoinkernel/commit/49b0bd0547b84fbe19f84518307ddeea6583da4b

Nice!

PR #104, opened 2025-11-10
Feat(core): Add Debug and Display implementations for Txid types

alexanderwiederin:
### Changes

Implements Display for Txid and TxidRef to show ids as hex strings in Bitcoin's standard format (reverse byte order).

Implements Debug for Txid and TxidRef in internal order.


TheCharlatan (approved):
lgtm

PR #109, opened 2025-11-10
Feat(core): Add `is_all_zeros()` and `is_null()` Methods to identify Coinbase outpoints

alexanderwiederin:
Add convenience methods for idenifying null outpoints and zero txids:

- TxidExt::is_all_zeros() checks if a txid consists of all zero bytes
- TxOutPointExt::isNull() identifies coinbase outpoints(index u32::MAX with all-zero txid)

Tests verify both owned and ref variants work correctly with coinbase transactions from test block data.

### DIFF
--- a/src/core/transaction.rs
+++ b/src/core/transaction.rs
@@ -845,6 +845,21 @@ mod tests {
         (tx1, tx2)
     }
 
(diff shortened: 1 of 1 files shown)


TheCharlatan (approved):
ACK 477e0774d043b43454b466c252c289ef97da2344

Nice :)

PR #112, opened 2025-11-11
Feat(core): Improve Documentation for Transaction module

alexanderwiederin:


src/core/transaction.rs
```diff
@@ -349,6 +845,32 @@ pub trait TxOutExt: AsPtr<btck_TransactionOutput> {
 ///
 /// Transaction outputs can be created from a script pubkey and amount, or retrieved
 /// from existing transactions. They represent spendable coins in the UTXO set.
+///
+/// # Structure
+///
+/// Each output contains:
+/// - **Value**: Amount in satoshis (1 BTC = 100,000,000 satoshis)
+/// - **Script pubkey**: The locking script that defines how the output can be spent
```
TheCharlatan: I think we can skip this section. Should be clear what it is  from the description preceding it.

src/core/transaction.rs
```diff
@@ -444,17 +997,58 @@ impl<'a> Clone for TxOutRef<'a> {
...
 
 /// A single transaction input referencing a previous output to be spent.
+///
+/// Transaction inputs specify which previous outputs are being spent. Each input contains:
+/// - **Outpoint**: Reference to the previous output (txid + index)
+/// - **Script signature**: Proof that the spender can satisfy the output's conditions
+/// - **Sequence**: Used for relative lock time and replace-by-fee
+///
+/// **Note**: The script signature and sequence fields are not currently accessible
+/// through this API. You can access the outpoint via [`outpoint`](TxInExt::outpoint).
```
TheCharlatan: I think we should just skip all of this. We can always add a complete description once all the getters are there.

PR #116, opened 2025-11-11
Refactor(core): Make ScriptVerifyStatus private

alexanderwiederin:
### Reasoning

ScriptVerifyStatus is an internal implementation detail used only to communicate status codes from the C verification function. Users interact with the public verify() function which returns Result<(), KernelError>, so this enum doesn't need to be exposed in the public API.


TheCharlatan (approved):
lgtm

PR #122, opened 2025-11-12
subtree: Update to latest master

sedited:
Had a small change in the kernel code, that should not have an effect here, but best to exercise it.


alexanderwiederin:
ACK https://github.com/TheCharlatan/rust-bitcoinkernel/pull/122/commits/fcf62b10cccabcc738acec5416433b8ab365fdee

PR #123, opened 2025-11-12
Feat(docs): add CI job to check for documentation warnings 

alexanderwiederin:
### Changes

- Fixes documentation warnings
- Adds CI job to check for documentation warnings


alexanderwiederin:
on second thought - we could add it to an existing job. 

@TheCharlatan  you are in a better position to judge.

TheCharlatan:
> on second thought - we could add it to an existing job.

Yes, I think this should go into the existing "Build and Test" job.

PR #125, opened 2025-11-24
fuzz: Extend block and transaction round trip tests

marcofleon:
Adds a new `kernel_primitives` fuzz target that tests the Rust wrapper around Bitcoin Core's block and transaction data structures.

Here's the [coverage report](https://marcofleon.github.io/coverage/kernel_primitives/).

### DIFF
--- a/fuzz/fuzz_targets/block_roundtrip.rs
+++ b/fuzz/fuzz_targets/block_roundtrip.rs
@@ -1,6 +1,6 @@
 #![no_main]
 
-use bitcoinkernel::Block;
+use bitcoinkernel::{prelude::*, Block, Transaction};
 use libfuzzer_sys::fuzz_target;
 
 fuzz_target!(|data: &[u8]| {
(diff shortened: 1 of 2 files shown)


fuzz/fuzz_targets/kernel_primitives.rs
```diff
@@ -0,0 +1,144 @@
...
+                    if let Ok(indexed_output) = tx.output(j) {
+                        assert_eq!(output.value(), indexed_output.value());
+                        assert_eq!(
+                            output.script_pubkey().to_bytes(),
+                            indexed_output.script_pubkey().to_bytes()
+                        );
+                    }
+                }
+            }
+        }
```
TheCharlatan: I think it would be nicer to extend the existing roundtrip targets with these additional tests. Both just consume a byte string and it would be good to have both of them consume a robust corpus of both hand-rolled block data and fuzzer inputs.

alexanderwiederin: Makes sense!

PR #140, opened 2026-02-23
Update subtree

sedited:
There have been some smaller cleanups and performance improvements (SpanReader, CCoinsCache changes) that would be good to get in here.


sedited:
Mmh, maybe we need to bump the msvc version in the CI akin to https://github.com/bitcoin/bitcoin/commit/7164a0cab650bdf01cdcbc3da690f6b674fcc7b3 .

sedited:
> nit: move the explicit windows version commit to before the libbitcoinkernel-sys changes.

Good idea, but will submit it as a separate PR so we get the CI actually running it.

alexanderwiederin (approved):

PR #148, opened 2026-04-06
fix(script_debug): emit DEBUG_SCRIPT frame for early-exit error paths

febyeji:
Three early-return paths in EvalScript exit before the DEBUG_SCRIPT call, so the debugger trace has no frame for the failing step:

  1. `SCRIPT_ERR_BAD_OPCODE` — GetOp() decode failure
  2. `SCRIPT_ERR_PUSH_SIZE` — push data > 520 bytes
  3. `SCRIPT_ERR_OP_COUNT` — >201 non-push opcodes

  Move the main DEBUG_SCRIPT call to right after GetOp() succeeds, and add a dedicated call with OP_INVALIDOPCODE for decode failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,8 @@ jobs:
       run: |
         # Ensure we're using the right version
         rustc --version
+        # Pin rustc-hash: 2.1.2+ requires rustc 1.77, incompatible with MSRV 1.71.
+        cargo update -p rustc-hash --precise 2.1.1
         # Build the project
         cargo build --all-features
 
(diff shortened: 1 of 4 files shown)


sedited (approved):
Makes sense to me. As long as we don't miss an actual state mutation (which we shouldn't as long as we call after `GetOp`, pushing the DEBUG_SCRIPT calls back a bit seems like the correct thing to do.

sedited:
Looks like this needs a rebase after your other PR has been merged.

sedited:
Thanks for picking up the CI patch, this looks ready to me now!

PR #149, opened 2026-04-06
feat(script_debug): add ScriptTrace and expose tapleaf/codesep fields

febyeji:
Expose `tapleaf_hash` and `codeseparator_pos` in the debug callback for tapscript sighash debugging, and add `ScriptTrace`, which is a high-level wrapper that captures a complete execution trace via `from_verify()` with indexed random-access to every frame.

### DIFF
--- a/examples/src/script_debug.rs
+++ b/examples/src/script_debug.rs
@@ -25,7 +25,7 @@ fn main() {
             _ => "UNKNOWN",
         };
         println!(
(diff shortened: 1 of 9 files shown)


sedited (approved):
LGTM

ScriptTrace seems like a useful data structure to expose, I would expect consuming clients to implement this themselves otherwise, so this could save some code. Adding some richer data is also good!

Issue #58, opened 2025-10-17
Refactor: implement Generic Iterator type


alexanderwiederin:
Due to MRSV 1.71 we could not implement a generic iterator type as suggested in https://github.com/TheCharlatan/rust-bitcoinkernel/pull/55.

Once [Floresta's](https://github.com/vinteumorg/Floresta) requirement of rust 1.74.1 is lifted, we can reimplement what was suggested in the noted PR.

PR #150, opened 2026-04-07
Add MSRV Cargo-minimal.lock and Cargo-recent.lock files

### LINKED ISSUE #147: Add MSRV Cargo.lock file
https://github.com/sedited/rust-bitcoinkernel/pull/136 is currently failing the MSRV check, which can happen if a dependency breaks it.  To avoid that, a Cargo.lock file can be committed which keeps a version of the dependencies in a locked state.  For example, rust-bitcoin keeps a copy of [Cargo-minimal.lock](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/Cargo-minimal.lock) and then uses that lock file in a MSRV CI test.

alexanderwiederin:
Fixes #147

### Summary
Without commited lock files, CI dependency resolution can pull in transitive dependencies that require a newer compiler than the project's MSRV. The triggering issue was that bindgen v0.72 depends on rustc-hash v2.x, and the recent release of rustc-hash v2.1.2 introduced a rustc 1.77.0 requirement, breaking the MSRV build against rustc 1.71.0.

### CI Changes
Cargo-minimal.lock pins dependencies to their oldest MSRV-compatible versions. Cargo-recent.lock pins to the latest patch versions compatible with the MSRV compiler. The CI msrv-check job now runs twice as a matrix, once with each lock file, similar to the approach used by rust-bitcoin.

#### libbitcoinkernel-sys
rustc-hash is pinned to below 2.1.1 to prevent bindgen from resolving to incompatible release. 2.1.0 in the minimal lockfile and 2.1.1 in the recent lockfile. 

#### dev-dependencies
tempdir is pinned to >=0.3.6 to prevent it from resolving to an older version that depends on an incompatible rand release under minimal-versions.

As a follow-up, I would suggest we remove tempdir as a dependency (https://github.com/sedited/rust-bitcoinkernel/issues/151)


.github/workflows/ci.yml
```diff
@@ -28,15 +32,18 @@ jobs:
     - name: Cache dependencies
       uses: Swatinem/rust-cache@v2
 
+    - name: Set dependencies
```
jaoleal:
```suggestion
    - name: Set Cargo.lock
```

jaoleal (approved):
ACK 1bfecb9d9e289ea7ecd4319d5a71bdd0b55d2047

Just a nit about the name of the step

.github/workflows/ci.yml
alexanderwiederin: Done - thanks!

jaoleal:
ACK da606fa

libbitcoinkernel-sys/Cargo.toml
```diff
@@ -53,3 +53,4 @@ publish = true
 [build-dependencies]
 cc = "1.2"
 bindgen = "0.72"
+rustc-hash = "=2.1.1"
```
yancyribbens: Won't this mean that if you do not need the MSRV, for exmple if you can run stable, then this will preven rustc-hash from using the latest version when it could?  If you need this version to work for the MSRV, then why not just defer to the lockfile?

alexanderwiederin: That is a good point.

The reason I pinned it is that a user running rust `v1.71.0` that calls `cargo update` will have a broken build unless we have this pin in `Cargo.toml` or ensure that they use the committed lock files. 

But thinking about it again, maybe a disclaimer for `1.71.0` users to add `--locked` might be the lesser evil. What do you think?

alexanderwiederin: In fact it's not just `v1.71.0` users that are affected but all users on rustc older than `1.77.0`

https://crates.io/crates/rustc-hash/2.1.2

alexanderwiederin: Opened an issue for rustc-hash: https://github.com/rust-lang/rustc-hash/issues/68

alexanderwiederin: We could also consider lifting the MSRV. If I remember correctly we set it to `1.71.0` for Floresta, but they seem to have bumped to `1.81.0` https://github.com/getfloresta/Floresta/pull/687

alexanderwiederin: Given the answer in https://github.com/rust-lang/rustc-hash/issues/68#issuecomment-4207073772, I would suggest we bump MSRV to `1.77.0` and remove the pin.

yancyribbens: Generally providing the lockfile for downstream users to reference and use is about the best that can be done in this situation.  It's really the fault of said crate which breaks the MSRV, which is why ideally you want as few deps as possible to avoid that situation.  It's a really easy mistake for crate owners to make though unfortunately.

If there's other reasons to upgrade to `1.77.0` that might make sense, although using the lockfile might be better to keep the MSRV low.

alexanderwiederin: We do have https://github.com/sedited/rust-bitcoinkernel/issues/58, which was blocked due to MSRV - not reason enough to bump on its own, but something to bear in mind. I also wonder if we can/should start to think about removing bindgen and write our own bindings, which has been discussed before.

Given all this, I think the best course of action is:
1. Implement the lockfile as suggested by @yancyribbens on this PR as an immediate fix for the CI
2. Open an issue to discuss MSRV and bindgen holistically

@sedited, will be offline for a few more weeks.

Any thoughts?

yancyribbens: The Cargo.toml file looks better now except for `tempdir`.  Did I see that it is going to be removed as a dep?  Maybe that should be done before this gets merged.

No opinion on bindgen since I don't know much of the background of the discussion.  Opening a PR to discuss sounds good though.

alexanderwiederin: Yes, the tempdir removal is being tracked in #153.

I would prefer to get the CI fix merged first and follow up with the `tempdir` removal.

yancyribbens: fair.  I think it would be a good idea to remove tempdir than before publishing the next release on crates.

yancyribbens: s/than/then

yancyribbens: Oh, tempdir is a dev dependency, so it doesn't matter so much.

alexanderwiederin: Thanks! If you feel comfortable, could leave an ACK? That will make @sedited`s life easier when he gets back.

README.md
```diff
@@ -34,6 +34,16 @@ dependencies. Once setup, run:
 cargo b
 ```
 
+## MSRV (Minimum Supported Rust Version)
+
+The minimum supported Rust version is 1.71.0. Users on rustc older than
```
yancyribbens: nit: change `1.71.0` to `1.71` here and in the following line.  The 0 is implicit so can be dropped.

libbitcoinkernel-sys/Cargo.toml
yancyribbens: Sure, I left a minor nit.  Am happy to ACK the current commit or an updated one.

yancyribbens:
utACK cd69fe283a4249720e3694c12f1e2d454ea19c08

jaoleal (approved):
ACK cd69fe283a4249720e3694c12f1e2d454ea19c08

luisschwab:
FIY: [`cargo-rbmt`](https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/tree/master/cargo-rbmt) is great to manage this kind of stuff. It's what `rust-bitcoin` uses.

alexanderwiederin:
> FIY: [`cargo-rbmt`](https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/tree/master/cargo-rbmt) is great to manage this kind of stuff. It's what `rust-bitcoin` uses.

Interesting! Will look into it. Thanks for the pointer.

PR #153, opened 2026-04-07
Chore: remove tempdir dependency

### LINKED ISSUE #151: Replace tempdir with std::env::temp_dir()
`tempdir` is an unmaintained crate that has not been updated in eight years. It is used in two places, both in tests:

- `tests/test.rs`
- `src/state/chainstate.rs`

This will allow us to remove the `tempdir = ">=0.3.6"` minimum version constraint that was added in #150.

jaoleal:
Fix #151.

This PR removes the [tempdir](https://github.com/rust-lang-deprecated/tempdir) crate that is deprecated.

The removal reflected some changes on tests as expected. 

The created path was also exported for those tests but they weren't used and I took the chance to remove those vars.


alexanderwiederin:
~I think this should build on https://github.com/sedited/rust-bitcoinkernel/pull/150 - just be aware that we may need to rebase.~

To replicate the behaviour of `tempdir` we also need to clean up the directories after the tests have run. To make it reusable, I would recommend adding a `TempDir` struct similar to:

```rust
pub struct TempDir {
    data_dir: std::path::PathBuf,
    blocks_dir: std::path::PathBuf,
}

impl TempDir {
    pub fn new(name: &str) -> Self {
        let data_dir = std::env::temp_dir().join(name);
        std::fs::create_dir_all(&data_dir).expect("failed to create temp dir");
        let blocks_dir = data_dir.join("blocks");
        Self { data_dir, blocks_dir }
    }

    pub fn data_dir(&self) -> &str {
        self.data_dir.to_str().expect("temp dir path is not valid UTF-8")
    }

    pub fn blocks_dir(&self) -> &str {
        self.blocks_dir.to_str().expect("temp dir path is not valid UTF-8")
    }
}

impl Drop for TempDir {
    fn drop(&mut self) {
        let _ = std::fs::remove_dir_all(&self.data_dir);
    }
}
```

To avoid exposing this in the public API, I think this struct is simple enough to be duplicated in `src/test_utils.rs` and `test/common.rs` files.

What do you think?

jaoleal:
>What do you think?

Cant we reach the same behavior with a versioning or numbering of the directory ?

alexanderwiederin:
> Cant we reach the same behavior with a versioning or numbering of the directory ?

I don't think so. Numbering/versioning would help with uniqueness, but that is a separate concern from cleanup.

jaoleal:
done, @alexanderwiederin!

src/ffi/test_utils.rs
```diff
@@ -102,6 +102,42 @@ macro_rules! test_ref_copy {
     };
 }
 
+/// Utility to create temporary directories that are cleaned on drop.
```
alexanderwiederin: I think the `ffi` directory should stay focused on `ffi` and would recommend we move this to a new file `src/test_utils.rs`, with the module declared as #[cfg(test)] in `lib.rs`.

src/ffi/test_utils.rs
```diff
@@ -102,6 +102,42 @@ macro_rules! test_ref_copy {
...
+pub struct TempDir {
+    data_dir: std::path::PathBuf,
+    blocks_dir: std::path::PathBuf,
+}
+
+impl TempDir {
+    pub fn new(name: &str) -> Self {
+        let data_dir = std::env::temp_dir().join(name);
+        std::fs::create_dir_all(&data_dir).expect("failed to create temp dir");
+        let blocks_dir = data_dir.join("blocks");
```
alexanderwiederin: I suggest we create the `blocks_dir` also. This will make `TempDir` self-contained and removes the implicit dependency on the chainstate manager to create it. 

```suggestion
        let blocks_dir = data_dir.join("blocks");
        std::fs::create_dir_all(&blocks_dir).expect("failed to create temp dirs");
```

src/ffi/test_utils.rs
alexanderwiederin: I would also note in the comments that this is duplicated in `test/common.rs`

tests/test.rs
```diff
@@ -15,14 +15,49 @@ mod tests {
...
+                .to_str()
+                .expect("temp dir path is not valid UTF-8")
+        }
+    }
+
+    impl Drop for TempDir {
+        fn drop(&mut self) {
+            let _ = std::fs::remove_dir_all(&self.data_dir);
+        }
+    }
```
alexanderwiederin: Would recommend moving this to a new file `test/common.rs` and adding to the comment section that this is duplicated in `src/test_utils.rs`

The feedback for the other `TempDir` applies here too.

src/ffi/test_utils.rs
```diff
@@ -102,6 +102,42 @@ macro_rules! test_ref_copy {
...
 
+/// Utility to create temporary directories that are cleaned on drop.
+pub struct TempDir {
+    data_dir: std::path::PathBuf,
+    blocks_dir: std::path::PathBuf,
+}
+
+impl TempDir {
+    pub fn new(name: &str) -> Self {
+        let data_dir = std::env::temp_dir().join(name);
```
alexanderwiederin: For uniqueness and to avoid collisions, I would do this:
```suggestion
        let id = COUNTER.fetch_add(1, Ordering::Relaxed);
        let data_dir = std::env::temp_dir().join(format!("{name}_{id}"));
```

With 
```rust 
static COUNTER: AtomicU64 = AtomicU64::new(0);
```
defined before the struct. 

`Ordering::Relaxed` is sufficient here because we only care that each call gets a unique number, not about the ordering of other memory operations around the increment.

tests/test.rs
```diff
@@ -111,12 +145,11 @@ mod tests {
 
     fn setup_chainman_with_blocks(
         context: &Arc<Context>,
-        data_dir: &str,
+        temp_dir: TempDir,
```
alexanderwiederin: I think this should take a reference:
```suggestion
        temp_dir: &TempDir,
```
This way the caller retains the ownership of `TempDir`. If not, I think the files will be dropped at the end of `setup_chainman_with_blocks`, deleting the directory while the returned `ChainstateManager` is still using it.

alexanderwiederin (changes requested):
Cool!

alexanderwiederin:
Also don't forget to remove `tempdir` from the `Cargo.toml` :)

src/ffi/test_utils.rs
jaoleal: Wait, i had placed the struct under `ffi/` because i did a find on the `test_uilts.rs` and it showed me that file. I created now `src/test_utils.rs` but i will have to create `test/common.rs` ? Two files with a single and duplicated struct is really weird...

It can be imported from `src/test_utils.rs`, is that fine ?

src/ffi/test_utils.rs
jaoleal: I like your magic tools magic man

tests/test.rs
jaoleal: https://github.com/sedited/rust-bitcoinkernel/pull/153#discussion_r3087568344

jaoleal:
Sorry, fighting the CI.

Okay, addressed suggestions on [30cb2ad](https://github.com/sedited/rust-bitcoinkernel/pull/153/commits/30cb2ad79e61a5e5b6bef7bc0c5b5e6d3560c9f1) , to avoid that duplication and repetition which could lead to code drift i preferred another approach similar to whats done on [bip39 rust crate](https://github.com/rust-bitcoin/rust-bip39/blob/d6dbc31678cc507c8cae62b3a059b0b48e866436/Cargo.toml#L57)

src/ffi/test_utils.rs
alexanderwiederin: I agree it's not ideal, but I want to avoid polluting the API surface with `TempDir` (even behind a feature gate). Since it's small, self-contained and unlikely to evolve, I think the cost of duplication is low.

alexanderwiederin:
> Okay, addressed suggestions on [30cb2ad](https://github.com/sedited/rust-bitcoinkernel/pull/153/commits/30cb2ad79e61a5e5b6bef7bc0c5b5e6d3560c9f1) , to avoid that duplication and repetition which could lead to code drift i preferred another approach similar to whats done on [bip39 rust crate](https://github.com/rust-bitcoin/rust-bip39/blob/d6dbc31678cc507c8cae62b3a059b0b48e866436/Cargo.toml#L57)

Appreciate you looking at rust-bip39. 

From what I understand `rand` is an optional dependency in `Cargo.toml` which is re-exported behind `#[cfg(feature = "rand")]` so the benches get access to a compatible version. 

I don't think that is quite what we need. Do you agree?

jaoleal:
>I don't think that is quite what we need. Do you agree?

I thought that importing itself as a dev dependency is weird too, agree with that... But that addresses a drift problem if this new struct thats being introduced were evolved or refined, avoiding a possible maintenance burn, it also sounds solid if new test components are needed. For me that was enough to suggest it here but i think is as valid as your suggestion, do you want for me to stick with it ?

src/test_utils.rs
```diff
@@ -0,0 +1,48 @@
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::{env, fs, path::PathBuf};
+
+/// A counter that is concatenated at the end of each `CMTempDir` instance to provide uniqueness and avoid collisions.
```
alexanderwiederin: If we go with duplication instead of exposing this via a feature gate, we can drop the documentation comments.

src/test_utils.rs
```diff
@@ -0,0 +1,48 @@
...
+static COUNTER: AtomicU64 = AtomicU64::new(0);
+
+/// Utility to create temporary directories with enough structure for a `ChainstateManager`, directories are cleaned on drop.
+pub struct CMTempDir {
```
alexanderwiederin: I think calling it `TempDir` is more natural.

alexanderwiederin (commented):
I appreciate the maintenance argument, but I think we should be conservative about what we expose as a feature. TempDir is test infrastructure and I'd rather not have it appear in the public API even behind a feature gate. Given how stable the struct is likely to be, I think the duplication cost is acceptable. I'd go with duplication.

jaoleal:
>I'd rather not have it appear in the public API even behind a feature gate

Yes, thats a considerable flaw.

Applied suggestions on 3c65b25

Plase take a look

tests/common.rs
```diff
@@ -0,0 +1,47 @@
+use std::sync::atomic::{AtomicU64, Ordering};
```
alexanderwiederin, line 1: I found this log running the tests:
```
test result: ok. 233 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.12s

     Running tests/common.rs (target/debug/deps/common-9c3bf7c061820654)
```

Which means that `tests/common.rs` is being compiled as its own test binary. From what I understand the conventional fix is to move the file to `tests/common/mod.rs`, which cargo won't compile as a standalone binary. 

Can we try that?

tests/common.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::{env, fs, path::PathBuf};
+
+static COUNTER: AtomicU64 = AtomicU64::new(0);
+
+/// Utility to create temporary directories that are cleaned on drop. Exact same struct as found in `src/test_utils.rs`.
```
alexanderwiederin, line 6: I think this should be a normal comment: 
```
// Utility to create temporary directories that are cleaned on drop. Duplicated in `src/test_utils.rs` and `tests/common/mod.rs`.
```
It's describing an implementation details for maintainers (the duplication), not documenting a public API.

Same applies to `src/test_utils.rs`.

tests/common.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct TempDir {
+    data_dir: PathBuf,
+    blocks_dir: PathBuf,
+}
+
+impl TempDir {
+    /// Creates a new `TempDir` from a given name concatenating it with an unique id.
```
alexanderwiederin, line 13: This can be removed along with line 28 and 35. It's self-explanatory. The less content the files have the smaller the risk of drift. 

Same applies to `src/test_utils.rs`.

alexanderwiederin (commented):
Thanks for addressing that!

I think the number of characters per line in the body of the commit message is too long. Please follow https://cbea.ms/git-commit/#wrap-72, which suggest a maximum character length of 72.

We should also add to the commit message why TempDir is duplicated. Something like: "TempDir is duplicated in `src/test_utils.rs` and `tests/common/mod.rs` to avoid exposing it in the public API."

tests/test.rs
```diff
@@ -131,15 +131,19 @@ mod tests {
...
             let block_data = read_block_data();
 
-            let chainman = ChainstateManagerBuilder::new(&context, &data_dir, &blocks_dir)
-                .unwrap()
-                .build()
-                .unwrap();
+            let chainman = ChainstateManagerBuilder::new(
+                &context,
+                &temp_dir.data_dir(),
+                &temp_dir.blocks_dir(),
```
alexanderwiederin:
```suggestion
                temp_dir.data_dir(),
                temp_dir.blocks_dir(),
```

src/test_utils.rs
```diff
@@ -0,0 +1,43 @@
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::{env, fs, path::PathBuf};
+
+static COUNTER: AtomicU64 = AtomicU64::new(0);
+
+// Utility to create temporary directories that are cleaned on drop. Duplicated in `src/test_utils.rs` and `tests/common/mod.rs`.pub struct TempDir {
```
alexanderwiederin: `pub struct TempDir {` was swallowed by the comment here.

alexanderwiederin (changes requested):
Almost there!

alexanderwiederin:
Perfect! 

The lock files appear to have been generated in the wrong order. Please run: 
```
cargo +nightly update -Z minimal-versions
cp Cargo.lock Cargo-minimal.lock

cargo update
cp Cargo.lock Cargo-recent.lock
``` 
And amend your commit with the result. 

I'll add these commands to the README and a CI check to catch this automatically going forward.

jaoleal:
Well i ran the commands and even tried to identify what went worng with the given commands but its confusing that even when running the recent pair with
```bash
$ cargo +1.71 update # 1.71 specified here
    Updating crates.io index
 $ cp Cargo.lock Cargo-recent.lock
```

I get that lock file thats erroring in the CI.

cc @alexanderwiederin

alexanderwiederin:
Thanks! 

The issue with `cargo +1.71 update` is that it uses the 1.71 toolchain to run the resolver, but that doesn't enforce MSRV constraints on transitive dependencies. 

Had forgotten this constraint when regenerating earlier. 

Instead, let's run the default toolchain but explicitly pin afterwards. 
```
cargo update
cargo update -p rustc-hash --precise 2.1.1
cp Cargo.lock Cargo-recent.lock
```
and amend the commit?

All the more reason to remove bindgen.

jaoleal:
Okay, that resolved the problem.

cc @alexanderwiederin, are we done here ?

alexanderwiederin (approved):
Looks good to me - thanks for hanging in there!

PR #158, opened 2026-04-10
Feat: Android support

### LINKED ISSUE #175: ci: add cross-compilation job for 32-bit/Android targets
https://github.com/sedited/rust-bitcoinkernel/pull/163#discussion_r3170645043

Formally supporting 32-bit targets should also be mentioned in the libbitcoinkernel-sys README

jaoleal:
This pr contain 3 commits:

1. Build instructions for android building, concise and aiming to be executed on the nix environment introduced in the second commit.
2. Nix expressions on flake.nix gathering dependencies for building and testing the android output.
3. CI runner that evaluates the nix expressions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,29 @@ jobs:
     - name: Build and test
       run: cargo test -vv
 
+
+  android:
+    name: Android (${{ matrix.package }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - libbitcoinkernel-android-aarch64
+          - libbitcoinkernel-android-armv7
+          # Build-only: Rust 1.71 compiler_builtins lacks f128 routines
+          # required by bionic on x86_64. Tests are skipped until the
+          # MSRV is bumped or a workaround is found. See #198
+          - libbitcoinkernel-android-x86_64
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install Nix
+      uses: cachix/install-nix-action@v31
+
+    - name: Build${{ matrix.package != 'libbitcoinkernel-android-x86_64' && ' and Test' || '' }}
+      run: nix build .#${{ matrix.package }} -L
+
   fuzz-corpus:
     name: Verify Fuzz Corpus
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added Nix package outputs for Android with bundled NDK r24, Rust toolchains, Boost, and cmake.
 - Added `Block::check` to perform context-free validation of a block (size, weight, coinbase, transactions, sigops), with optional proof-of-work and merkle-root checks toggled via the `BLOCK_CHECK_BASE` / `_POW` / `_MERKLE` / `_ALL` flags. Returns a `BlockCheckResult` enum carrying the validation state on failure.
 - Implemented `Debug` for `BlockValidationResult`, `BlockValidationStateRef`, `ProcessBlockHeaderresult` and `BlockCheckResult`, enabling inspection via `{:?}` in logs and test output.
 - Added `#[must_use]` to `BlockCheckResult` and `ProcessBlockHeaderResult` to warn when validation results are silently ignored.
@@ -36,7 +37,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `From<BlockHeader>` and `From<&BlockHeader>` for `[u8; 80]` to align with existing block conversions.
 
 ### Dependencies
-
 - The sys crate no longer uses auto-generated bindings from bindgen. This removes some build-time dependencies for this crate.
 
 ## [0.2.0] 2026-01-26
--- a/README.md
+++ b/README.md
@@ -34,6 +34,23 @@ dependencies. Once setup, run:
 cargo b
 ```
 
+### Android Cross-Compilation
+
+Android cross-compilation requires [Nix](https://nixos.org/).
+
+Nix handles the exact NDK version, Rust toolchains, Boost, and cmake
+automatically, giving you a reproducible build environment with no manual setup.
+
+```bash
+nix build .#libbitcoinkernel-android-aarch64
+nix build .#libbitcoinkernel-android-armv7
+nix build .#libbitcoinkernel-android-x86_64 # build only; tests skipped (see flake.nix)
+```
+
+The resulting libraries and headers are placed in `result/lib/` and `result/include`.
+
+Output targets Android API 24+ (Nougat) minimum.
+
 ## MSRV (Minimum Supported Rust Version)
 
 The minimum supported Rust version is 1.71. Users on rustc older than
--- a/flake.nix
+++ b/flake.nix
@@ -25,9 +25,10 @@
         };
 
         rustVersion = "1.71.0";
+        rustToolchainSha256 = "sha256-ks0nMEGGXKrHnfv4Fku+vhQ7gx76ruv6Ij4fKZR3l78=";
         rustToolchain = fenix.packages.${system}.fromToolchainName {
           name = rustVersion;
-          sha256 = "sha256-ks0nMEGGXKrHnfv4Fku+vhQ7gx76ruv6Ij4fKZR3l78=";
+          sha256 = rustToolchainSha256;
         };
         rustBuildToolchain = fenix.packages.${system}.combine [
           rustToolchain.rustc
@@ -85,6 +86,129 @@
             pkgs.gcc.cc.lib
           ];
         };
+
+        packages =
+          # Android build infrastructure (unfree NDK + SDK).
+          let
+            ndkVersion = "27.2.12479018"; # which NDK release to download
+            lockfile = ./Cargo-minimal.lock;
+            ANDROID_API_LEVEL = "24";
+            crateVersion =
+              (builtins.fromTOML (builtins.readFile ./libbitcoinkernel-sys/Cargo.toml)).package.version;
+
+            androidPkgs = import nixpkgs {
+              inherit system;
+              config.android_sdk.accept_license = true;
+              config.allowUnfree = true;
+            };
+            androidComposition = androidPkgs.androidenv.composeAndroidPackages {
+              # platformVersions is the SDK tooling version, not the minimum API level.
+              # The NDK target floor is set via ANDROID_API_LEVEL in build.rs (default 24).
+              platformVersions = [ "34" ];
+              ndkVersions = [ ndkVersion ];
+              includeNDK = true;
+            };
+            androidSdk = androidComposition.androidsdk;
+            androidNdk = "${androidSdk}/libexec/android-sdk/ndk/${ndkVersion}";
+
+            runTests = system == "x86_64-linux";
+
+            # x86_64-linux-android is excluded from tests: Rust 1.71's
+            # compiler_builtins lacks the f128 routines (__eqtf2, __multf3, …)
+            # that bionic's static libc.a requires on x86_64.
+            #
+            # See #198
+            testableTargets = [
+              "aarch64-linux-android"
+              "armv7-linux-androideabi"
+            ];
+
+            mkAndroidPackage =
+              rustTarget:
+              let
+                canTest = runTests && builtins.elem rustTarget testableTargets;
+                rustTargetToolchain = fenix.packages.${system}.combine [
+                  rustToolchain.rustc
+                  rustToolchain.cargo
+                  rustToolchain.rust-src
+                  rustToolchain.rust-std
+                  (fenix.packages.${system}.targets.${rustTarget}.fromToolchainName {
+                    name = rustVersion;
+                    sha256 = rustToolchainSha256;
+                  }).rust-std
+                ];
+                rustPlatform = androidPkgs.makeRustPlatform {
+                  cargo = rustTargetToolchain;
+                  rustc = rustTargetToolchain;
+                };
+                qemuBin =
+                  if builtins.match "aarch64.*" rustTarget != null then
+                    "qemu-aarch64"
+                  else if builtins.match "armv7.*" rustTarget != null then
+                    "qemu-arm"
+                  else
+                    throw "Unsupported Android target: ${rustTarget}";
+                # NDK clang wrapper triple: armv7 uses "armv7a-linux-androideabi",
+                # all others match the Rust target triple.
+                ndkClangTriple =
+                  if builtins.match "armv7.*" rustTarget != null then "armv7a-linux-androideabi" else rustTarget;
+                ndkLinker = "${androidNdk}/toolchains/llvm/prebuilt/linux-x86_64/bin/${ndkClangTriple}${ANDROID_API_LEVEL}-clang";
+                cargoTargetPrefix = "CARGO_TARGET_${
+                  builtins.replaceStrings [ "-" ] [ "_" ] (pkgs.lib.toUpper rustTarget)
+                }";
+                cargoRunnerEnvVar = "${cargoTargetPrefix}_RUNNER";
+                cargoLinkerEnvVar = "${cargoTargetPrefix}_LINKER";
+              in
+              rustPlatform.buildRustPackage {
+                pname = "libbitcoinkernel-${rustTarget}";
+                version = crateVersion;
+                src = ./.;
+                cargoLock.lockFile = lockfile;
+                postPatch = ''
+                  cp ${lockfile} Cargo.lock
+                '';
+                nativeBuildInputs = [
+                  androidPkgs.cmake
+                  androidPkgs.boost.dev
+                  androidSdk
+                ]
+                ++ pkgs.lib.optionals canTest [
+                  pkgs.qemu
+                ];
+
+                ANDROID_HOME = "${androidSdk}/libexec/android-sdk";
+                ANDROID_NDK_HOME = androidNdk;
+                CMAKE_PREFIX_PATH = "${androidPkgs.boost.dev}";
+
+                # cargoBuildHook hardcodes the host --target at
+                # derivation time, so we bypass it for cross builds.
+                dontCargoBuild = true;
+                doCheck = canTest;
+                buildPhase = ''
+                  cargo build -p libbitcoinkernel-sys --target ${rustTarget} --offline --release
+                '';
+                checkPhase = pkgs.lib.optionalString canTest ''
+                  export ${cargoLinkerEnvVar}=${ndkLinker}
+                  export ${cargoRunnerEnvVar}=${pkgs.qemu}/bin/${qemuBin}
+                  export QEMU_LD_PREFIX=${androidNdk}/toolchains/llvm/prebuilt/linux-x86_64/sysroot
+                  export RUSTFLAGS="-C target-feature=+crt-static"
+                  cargo test --target ${rustTarget} --offline --release --verbose
+                '';
+                installPhase = ''
+                  mkdir -p $out/lib $out/include
+                  find target/${rustTarget}/release -path "*/out/install/lib/*.a" \
+                    -exec cp {} $out/lib/ \;
+                  find target/${rustTarget}/release -path "*/out/install/include/*" \
+                    -exec cp {} $out/include/ \;
+                '';
+              };
+          in
+          {
+            libbitcoinkernel-android-aarch64 = mkAndroidPackage "aarch64-linux-android";
+            libbitcoinkernel-android-armv7 = mkAndroidPackage "armv7-linux-androideabi";
+            libbitcoinkernel-android-x86_64 = mkAndroidPackage "x86_64-linux-android";
+            # i686 omitted: not a current target
+          };
       }
     );
 }
--- a/libbitcoinkernel-sys/CHANGELOG.md
+++ b/libbitcoinkernel-sys/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Added build support for Android
+
 ### Added
 - New `btck_TxValidationState` opaque type for holding transaction validation state
 - New `btck_TxValidationResult` type with named constants for all transaction validation result variants
--- a/libbitcoinkernel-sys/build.rs
+++ b/libbitcoinkernel-sys/build.rs
@@ -3,6 +3,18 @@ use std::path::Path;
 use std::process::Command;
 
 fn main() {
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+    let is_android = target_os == "android";
+
+    let ndk = if is_android {
+        Some(
+            env::var("ANDROID_NDK_HOME")
+                .expect("Android target detected but ANDROID_NDK_HOME is not set"),
+        )
+    } else {
+        None
+    };
+
     let bitcoin_dir = Path::new("bitcoin");
     let out_dir = env::var("OUT_DIR").unwrap();
     let build_dir = Path::new(&out_dir).join("bitcoin");
@@ -15,7 +27,8 @@ fn main() {
 
     let build_config = "RelWithDebInfo";
 
-    Command::new("cmake")
+    let mut cmake_configure = Command::new("cmake");
+    cmake_configure
         .arg("-B")
         .arg(&build_dir)
         .arg("-S")
@@ -38,7 +51,46 @@ fn main() {
         .arg("-DBUILD_SHARED_LIBS=OFF")
         .arg("-DCMAKE_INSTALL_LIBDIR=lib")
         .arg("-DENABLE_IPC=OFF")
-        .arg(format!("-DCMAKE_INSTALL_PREFIX={}", install_dir.display()))
+        .arg(format!("-DCMAKE_INSTALL_PREFIX={}", install_dir.display()));
+
+    if is_android {
+        let ndk = ndk.as_deref().unwrap();
+        let toolchain_file = format!("{ndk}/build/cmake/android.toolchain.cmake");
+
+        // Rust target triple -> NDK ABI name.
+        let abi = match env::var("TARGET").unwrap() {
+            t if t.contains("aarch64") => "arm64-v8a",
+            t if t.contains("armv7") => "armeabi-v7a",
+            t if t.contains("x86_64") => "x86_64",
+            target => panic!("Unsupported Android ABI: {target}"),
+        };
+
+        // API level 24+ is required because Bitcoin Core uses getifaddrs
+        // which was introduced in Android API 24 (Nougat).
+        //
+        // This can be overridden to a higher level by setting ANDROID_API_LEVEL.
+        let api_level = match env::var("ANDROID_API_LEVEL") {
+            Ok(level) => {
+                let n: u32 = level.parse().expect("ANDROID_API_LEVEL must be a number");
+                assert!(n >= 24, "ANDROID_API_LEVEL must be 24+");
+                level
+            }
+            _ => "24".to_string(),
+        };
+
+        cmake_configure
+            .arg(format!("-DCMAKE_TOOLCHAIN_FILE={toolchain_file}"))
+            .arg(format!("-DANDROID_ABI={abi}"))
+            .arg(format!("-DANDROID_PLATFORM=android-{api_level}"))
+            .arg("-DCMAKE_SYSTEM_NAME=Android")
+            .arg(format!("-DCMAKE_ANDROID_NDK={ndk}"))
+            // The Android NDK toolchain sets CMAKE_FIND_ROOT_PATH_MODE_PACKAGE
+            // to ONLY, which prevents cmake from finding host packages via
+            // CMAKE_PREFIX_PATH. Override it so Boost headers can be located.
+            .arg("-DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=BOTH");
+    }
+
+    cmake_configure
         .status()
         .expect("cmake should be installed and available in PATH");
 
@@ -70,19 +122,60 @@ fn main() {
     } else {
         install_dir.join("lib")
     };
+
     println!("cargo:rustc-link-search=native={}", lib_dir.display());
 
     println!("cargo:rustc-link-lib=static=bitcoinkernel");
 
     let compiler = cc::Build::new().get_compiler();
-    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
 
     if target_os == "windows" {
         println!("cargo:rustc-link-lib=bcrypt");
         println!("cargo:rustc-link-lib=shell32");
-    }
+    } else if is_android {
+        // Without these, changing either variable between builds won't trigger a rebuild,
+        // so we could silently end up with a stale binary built against the wrong NDK or API level.
+        println!("cargo:rerun-if-env-changed=ANDROID_NDK_HOME");
+        println!("cargo:rerun-if-env-changed=ANDROID_API_LEVEL");
+        // Android NDK ships libc++_static.a and libc++abi.a in the
+        // per-architecture sysroot directory (not the API-level subdirectory).
+        let ndk = ndk.as_deref().unwrap();
 
-    if compiler.is_like_clang() {
+        // Rust target triple -> NDK sysroot lib directory triple.
+        // armv7 differs: Rust says `armv7-linux-androideabi`, NDK says `arm-linux-androideabi`.
+        let ndk_triple = env::var("TARGET")
+            .map(|target| {
+                if target.starts_with("armv7") {
+                    "arm-linux-androideabi".into()
+                } else {
+                    target
+                }
+            })
+            .unwrap();
+
+        let host_tag = match std::env::consts::OS {
+            "macos" => "darwin-x86_64",
+            "linux" => "linux-x86_64",
+            os => panic!("unsupported build host for Android cross-compilation: {os}"),
+        };
+
+        let ndk_lib_dir =
+            format!("{ndk}/toolchains/llvm/prebuilt/{host_tag}/sysroot/usr/lib/{ndk_triple}");
+        println!("cargo:rustc-link-search=native={ndk_lib_dir}");
+        println!("cargo:rustc-link-lib=static=c++_static");
+        println!("cargo:rustc-link-lib=static=c++abi");
+
+        // The pre-compiled libcompiler_builtins for armv7-linux-androideabi
+        // ships ARM EABI helper symbols tagged with @@LIBC_N (e.g.
+        // __aeabi_memcpy@@LIBC_N).  When lld links a shared library or
+        // executable it errors because the LIBC_N version node is not
+        // defined in any version script.  --exclude-libs,ALL marks every
+        // symbol pulled from static archives as local, which suppresses
+        // the version-node error.
+        if ndk_triple == "arm-linux-androideabi" {
+            println!("cargo:rustc-link-arg=-Wl,--exclude-libs,ALL");
+        }
+    } else if compiler.is_like_clang() {
         if target_os == "macos" {
             println!("cargo:rustc-link-lib=dylib=c++");
         } else {
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 Cargo.lock
 coverage_report
+result


sedited:
This looks nice, can you rebase this?

jaoleal:
Changelog:

- Removed some PoC code
- Included android native building in ci (whitout nix).
- A fix regarding building on macos was included to.
- Rebased on top of #163 and included a fix regarding pointer sizes.

jaoleal:
I reduced a lot the code and i think this is the concise version to support android

alexanderwiederin:
~I don't think android is a supported target for libbitcoinkernel. Doesn't this require upstream work on bitcoin core first?~

flake.nix
```diff
@@ -35,6 +37,33 @@
...
+					(fenix.packages.${system}.targets.armv7-linux-androideabi.fromToolchainName {
+						name = rustVersion;
+						sha256 = "sha256-ks0nMEGGXKrHnfv4Fku+vhQ7gx76ruv6Ij4fKZR3l78=";
+					}).rust-std
+					(fenix.packages.${system}.targets.x86_64-linux-android.fromToolchainName {
+						name = rustVersion;
+						sha256 = "sha256-ks0nMEGGXKrHnfv4Fku+vhQ7gx76ruv6Ij4fKZR3l78=";
+					}).rust-std
+				];
+
```
alexanderwiederin: Can we contain the android configurations in devShells.android? I think the same applies to lines 18 and 19.

alexanderwiederin (commented):
From what I understand the ubuntu boost packages don't ship with the cmake config that core needs when cross-compiling. 

Am I right that you got it working on nix locally? If so, can we stick with with only supporting the nix build for android?

flake.nix
```diff
@@ -35,6 +37,33 @@
...
+				androidNdk = "${androidSdk}/libexec/android-sdk/ndk/27.2.12479018";
+
+				rustAndroidToolchain = fenix.packages.${system}.combine [
+					rustToolchain.rustc
+					rustToolchain.cargo
+					rustToolchain.rust-src
+					rustToolchain.rust-std
+					(fenix.packages.${system}.targets.aarch64-linux-android.fromToolchainName {
+						name = rustVersion;
+						sha256 = "sha256-ks0nMEGGXKrHnfv4Fku+vhQ7gx76ruv6Ij4fKZR3l78=";
```
alexanderwiederin: I noticed that the sha256 for all targets are the same. Shouldn't each have a different one because they are different binaries?

jaoleal: Ill check that

jaoleal:
> Am I right that you got it working on nix locally? If so, can we stick with with only supporting the nix build for android?

~~Yes, i worked all this on nix but Im not sure about leaving android building exclusively for nix, this would make it a direct dependency of the project and its not something really special that only nix can do. Besides the dependency problem, nix can drift versions because of nixpkgs releases and that can raise problems only by itself.~~

~~Therefore, i still would like for CI to keep the runner as ubuntu. Its easier for the general developer audience to replicate the build on their machine if they need~~

Edit: see the comment below

jaoleal:
Correction, I changed my mind... The idea is that the `build.rs` file is already bloated because of the android support additions, adding support for ubuntu bloated it even more... It looks like leaving to nix to cross-compile to android is actually a good choice, the changes stay minimal on `build.rs` side and it have broader support running on linux and macos machines... Ill make sure to document that somewhere and state how to make the android build.

jaoleal:
Okay the android build being exclusive for nix made things easier for this PR, the changes on build.rs are minimal... I included some inline doc comments so the reviewer can follow why each line were added.

I added a section on the readme explaining the android build but its not that explanatory in a level that one can learn how to build this lib to android, do you guys think thats needed ?

 I can extend the docs to include instructions for one to reproduce it on a downstream `build.rs`, just tell me where to put it.
AFAIK is trivial for those using Kotlin, as mandacaru is using, to pull an already compiled shared object during build time. Thats what ill do on floresta-nix for mandacaru, ideally the CI would also offer the compiled objects too for those who might need it.

libbitcoinkernel-sys/build.rs
```diff
@@ -40,7 +70,42 @@ fn main() {
...
+
+    if target_os == "android" {
+        let ndk =
+            android_ndk_home().expect("Android target detected but ANDROID_NDK_HOME is not set");
+        let toolchain_file = format!("{ndk}/build/cmake/android.toolchain.cmake");
+        assert!(
+            Path::new(&toolchain_file).exists(),
+            "Android NDK toolchain file not found at {toolchain_file}. \
+             Check that ANDROID_NDK_HOME points to a valid NDK installation"
+        );
```
alexanderwiederin: I think asserts are for faults in logic, not environment variables. I think a `!path.exists() { panic!(...) }` or `.expect()` would read more naturally here.

libbitcoinkernel-sys/build.rs
```diff
@@ -40,7 +70,42 @@ fn main() {
...
+        // which was introduced in Android API 24 (Nougat).
+        let api_level = env::var("ANDROID_API_LEVEL").unwrap_or_else(|_| "24".to_string());
+
+        cmake_configure
+            .arg(format!("-DCMAKE_TOOLCHAIN_FILE={toolchain_file}"))
+            .arg(format!("-DANDROID_ABI={abi}"))
+            .arg(format!("-DANDROID_PLATFORM=android-{api_level}"))
+            .arg("-DCMAKE_SYSTEM_NAME=Android")
+            .arg(format!("-DCMAKE_ANDROID_ARCH_ABI={abi}"))
+            .arg(format!("-DCMAKE_SYSTEM_VERSION={api_level}"))
```
alexanderwiederin: Do we really have to pass these in a second time after lines 96 and 97?

flake.nix
```diff
@@ -77,6 +77,81 @@
...
+					# Android build infrastructure (unfree NDK + SDK).
+					let
+						androidPkgs = import nixpkgs {
+							inherit system;
+							config.android_sdk.accept_license = true;
+							config.allowUnfree = true;
+						};
+						androidComposition = androidPkgs.androidenv.composeAndroidPackages {
+							platformVersions = [ "34" ];
+							ndkVersions = [ "27.2.12479018" ];
```
alexanderwiederin: `27.2.12479018` is duplicated on line 94. I would suggest we make it a variable.

flake.nix
```diff
@@ -77,6 +77,81 @@
...
+							rustTarget:
+							let
+								rustTargetToolchain = fenix.packages.${system}.combine [
+									rustToolchain.rustc
+									rustToolchain.cargo
+									rustToolchain.rust-src
+									rustToolchain.rust-std
+									(fenix.packages.${system}.targets.${rustTarget}.fromToolchainName {
+										name = rustVersion;
+										sha256 = "sha256-ks0nMEGGXKrHnfv4Fku+vhQ7gx76ruv6Ij4fKZR3l78=";
```
alexanderwiederin: The `sha256` here is already defined at the top of the flake as part of rustToolchain. Extract it into a `rustSha256` variable alongside `rustVersion` and reference it in both places. Otherwise a toolchain bump requires updating two identical hashes.

flake.nix
```diff
@@ -77,6 +77,81 @@
...
+				packages =
+					# Android build infrastructure (unfree NDK + SDK).
+					let
+						androidPkgs = import nixpkgs {
+							inherit system;
+							config.android_sdk.accept_license = true;
+							config.allowUnfree = true;
+						};
+						androidComposition = androidPkgs.androidenv.composeAndroidPackages {
+							platformVersions = [ "34" ];
```
alexanderwiederin: Can we add a comment above this? 

```
# platformVersions is the SDK tooling version, not the minimum API level.
# The NDK target floor is set via ANDROID_API_LEVEL in build.rs (default 24).
```

flake.nix
```diff
@@ -77,6 +77,81 @@
...
+																					-exec cp {} $out/lib/ \;
+																				find target/${rustTarget}/release -path "*/out/install/include/*" \
+																					-exec cp {} $out/include/ \;
+																			'';
+							};
+					in
+					{
+						android-aarch64 = mkAndroidPackage "aarch64-linux-android";
+						android-armv7 = mkAndroidPackage "armv7-linux-androideabi";
+						android-x86_64 = mkAndroidPackage "x86_64-linux-android";
```
alexanderwiederin: in `build.rs`, the `android_abi()` function handles four targets, but here only three are exposed. Is that intentional?

.github/workflows/ci.yml
```diff
@@ -211,6 +211,26 @@ jobs:
     - name: Build and test
       run: cargo test -vv
 
+
+  android-cross-compile:
+    name: Cross-Compile for Android (${{ matrix.package }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
alexanderwiederin, line 226: good call!

README.md
```diff
@@ -34,6 +34,22 @@ dependencies. Once setup, run:
 cargo b
 ```
 
+### Android Cross-Compilation
+
+Android cross-compilation requires [Nix](https://nixos.org/).
+
+Nix package outputs bundle the exact NDK version, Rust toolchains with Android
+targets, Boost, and cmake in a single reproducible build drying out the support
```
alexanderwiederin: what do you mean "drying" out the support?

README.md
```diff
@@ -34,6 +34,22 @@ dependencies. Once setup, run:
 cargo b
 ```
 
+### Android Cross-Compilation
```
alexanderwiederin, line 37: Worth adding a note that the output targets Android API 24+ (Nougat) minimum, so people know the minimum Android version their app must support.

alexanderwiederin (commented):
Great to see this working. 

If we land the hand-written bindings PR (https://github.com/sedited/rust-bitcoinkernel/pull/163) first, the changes in `build.rs` should become simpler. I would suggest we wait for it.

sedited:
@jaoleal I'm interested in getting this change through. Can you rebase it and address Alex' comments?

jaoleal:
> @jaoleal I'm interested in getting this change through. Can you rebase it and address Alex' comments?

Yes, sorry for taking long... I was waiting for the bindgen pr to be merged so we had sure what that would imply here without rebasing on a pr

.github/workflows/ci.yml
jaoleal: These new CI builds, they arent addressing #175 ?

jaoleal: If im missing a target let me know

alexanderwiederin: Yes, you are right it does. I have linked this PR to the issue.

jaoleal:
Addressed suggestions @sedited @alexanderwiederin

flake.nix
```diff
@@ -85,6 +86,85 @@
...
+                buildPhase = ''
+                        					cargo build -p libbitcoinkernel-sys --target ${rustTarget} --offline --release
+                  					'';
+                installPhase = ''
+                                    mkdir -p $out/lib $out/include
+                                    find target/${rustTarget}/release -path "*/out/install/lib/*.a" \
+                                    -exec cp {} $out/lib/ \;
+                                    find target/${rustTarget}/release -path "*/out/install/include/*" \
+                                    -exec cp {} $out/include/ \;
+                  							'';
```
sedited: What's going on with the indentation here?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.2.1] 2026-05-20
 
 ### Added
+- Added Nix package outputs for Android with bundled NDK r27, Rust toolchains, Boost, and cmake.
```
sedited: This needs to go in a new `###Added` entry in the unreleased section.

sedited:
@jaoleal how are you testing the build artifacts? Are you running the tests in some emulated environment, or do you just link the kernel crate into an existing project and observe that there are no failures?

flake.nix
jaoleal: Im not certain too, i tried to make it look normal but nixfmt 25.05 kept it like this, Ill check if this is indeed the correct indentation

jaoleal:
> @jaoleal how are you testing the build artifacts? Are you running the tests in some emulated environment, or do you just link the kernel crate into an existing project and observe that there are no failures?

None of these, I got satisfied by the build succeeding since testing this can be tricky... Are you looking into CI tests for this ?

But youre correct, we should be testing the outputs.

Testing it using floresta will be hard but ill look into it.

sedited:
Ah, I thought this was intended for eventual floresta integration. Or what was your motivation for working on it?

jaoleal:
>Ah, I thought this was intended for eventual floresta integration. Or what was your motivation for working on it?

Well, yes, im fixing a "floresta-issue" here but allow me to expose the whole picture.

Floresta uses bitcoin-kernel only for script validation, whitout the bitcoin-kernel feature, floresta just skips script validation.

Mandacaru, basically a port of floresta to android, needs floresta running on android, and is currently skipping script validation.

To test this using floresta i would need to: 1. Make floresta use this current `android_support` branch. 2. Build a binary and ship it to Mandacaru. 3. Build Mandacu, run it on android and perform full sync.

And this would not result in a satisfatory test because floresta would only exercise script validation. Too hard for basically useless results. 

AFAIK the best solution to be sure that bitcoin-kernel is indeed working on android is by adding a test in CI running inside an emulated environment, ill push it and lets see how it goes

sedited:
I'm also fine with just leaving it be for now and handling that in a follow-up pull request. Having something link is already a good assurance :)

jaoleal:
Ok, done.

32115fc Addressed suggestions:

- Moved changelog edits to unreleased section.
- Added tests for the android outputs.
- fixed formatting on string literals

cc @sedited take a look on https://github.com/sedited/rust-bitcoinkernel/actions/runs/26413972282/job/77754301637#step:4:6483

libbitcoinkernel-sys/build.rs
```diff
@@ -38,7 +59,38 @@ fn main() {
...
         .arg("-DCMAKE_INSTALL_LIBDIR=lib")
         .arg("-DENABLE_IPC=OFF")
-        .arg(format!("-DCMAKE_INSTALL_PREFIX={}", install_dir.display()))
+        .arg(format!("-DCMAKE_INSTALL_PREFIX={}", install_dir.display()));
+
+    let target = env::var("TARGET").unwrap();
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+
+    if target_os == "android" {
+        let ndk = env::var("ANDROID_NDK_HOME")
```
alexanderwiederin: We are fetching this twice. Any chance we can reuse this?

libbitcoinkernel-sys/build.rs
```diff
@@ -70,19 +122,34 @@ fn main() {
...
+        let ndk = env::var("ANDROID_NDK_HOME").expect("We called ANDROID_NDK_HOME before.");
 
-    if compiler.is_like_clang() {
+        let ndk_triple = android_sysroot_triple(&target);
+
+        let host_tag = if cfg!(target_os = "macos") {
+            "darwin-x86_64"
+        } else {
+            "linux-x86_64"
+        };
```
alexanderwiederin: Can we a helper next to `android_sysroot_triple`?

```rust
fn android_host_tag() -> &'static str {
    match std::env::consts::OS {
        "macos" => "darwin-x86_64",
        "linux" => "linux-x86_64",
        os => panic!("unsupported build host for Android cross-compilation: {os}"),
    }
}
```

libbitcoinkernel-sys/build.rs
```diff
@@ -2,6 +2,26 @@ use std::env;
...
+    match target {
+        t if t.contains("aarch64") => Some("arm64-v8a"),
+        t if t.contains("armv7") => Some("armeabi-v7a"),
+        t if t.contains("x86_64") => Some("x86_64"),
+        _ => None,
+    }
+}
+
+/// Rust target triple -> NDK sysroot lib directory triple.
+/// armv7 differs: Rust says `armv7-linux-androideabi`, NDK says `arm-linux-androideabi`.
```
alexanderwiederin: This can be a normal comment.

flake.nix
```diff
@@ -85,6 +86,125 @@
             pkgs.gcc.cc.lib
           ];
         };
+
+        packages =
+          # Android build infrastructure (unfree NDK + SDK).
+          let
+            ndkVersion = "27.2.12479018";
+            lockfile = ./Cargo-minimal.lock;
```
alexanderwiederin, line 94: Can we add this below: 
```nix
crateVersion = (builtins.fromTOML (builtins.readFile ./libbitcoinkernel-sys/Cargo.toml)).package.version;
```
and then use it on line 160?

libbitcoinkernel-sys/build.rs
```diff
@@ -2,6 +2,26 @@ use std::env;
 use std::path::Path;
 use std::process::Command;
 
+/// Rust target triple -> NDK ABI name (`arm64-v8a`, `armeabi-v7a`, …).
```
alexanderwiederin: This can be a normal `//` comment.

CHANGELOG.md
alexanderwiederin: This can be removed from the `## [0.2.1] 2026-05-20` section.

libbitcoinkernel-sys/build.rs
```diff
@@ -38,7 +59,38 @@ fn main() {
...
+        // which was introduced in Android API 24 (Nougat).
+        let api_level = "24";
+
+        cmake_configure
+            .arg(format!("-DCMAKE_TOOLCHAIN_FILE={toolchain_file}"))
+            .arg(format!("-DANDROID_ABI={abi}"))
+            .arg(format!("-DANDROID_PLATFORM=android-{api_level}"))
+            .arg("-DCMAKE_SYSTEM_NAME=Android")
+            .arg(format!("-DCMAKE_ANDROID_ARCH_ABI={abi}"))
+            .arg(format!("-DCMAKE_SYSTEM_VERSION={api_level}"))
```
alexanderwiederin: Raising this from earlier: do we really need to pass `abi` and `api_level` twice?

jaoleal:
Applied @alexanderwiederin suggestions

libbitcoinkernel-sys/build.rs
```diff
@@ -2,6 +2,34 @@ use std::env;
...
 
+// Rust target triple -> NDK ABI name (`arm64-v8a`, `armeabi-v7a`, …).
+fn android_abi(target: &str) -> Option<&'static str> {
+    match target {
+        t if t.contains("aarch64") => Some("arm64-v8a"),
+        t if t.contains("armv7") => Some("armeabi-v7a"),
+        t if t.contains("x86_64") => Some("x86_64"),
+        _ => None,
+    }
+}
```
alexanderwiederin: I think we have to be explicit about `i686`. Either support it or `panic!("unsupported Android ABI: {target}")`

libbitcoinkernel-sys/build.rs
```diff
@@ -38,7 +67,37 @@ fn main() {
...
+        let ndk = env::var("ANDROID_NDK_HOME")
+            .expect("Android target detected but ANDROID_NDK_HOME is not set");
+        let toolchain_file = format!("{ndk}/build/cmake/android.toolchain.cmake");
+
+        let abi =
+            android_abi(&target).unwrap_or_else(|| panic!("unsupported Android target: {target}"));
+
+        // API level 24+ is required because Bitcoin Core uses getifaddrs
+        // which was introduced in Android API 24 (Nougat).
+        let api_level = "24";
```
alexanderwiederin: How can people use higher levels of the API?

flake.nix
```diff
@@ -85,6 +86,88 @@
...
+                '';
+                nativeBuildInputs = [
+                  androidPkgs.cmake
+                  androidPkgs.boost.dev
+                  androidSdk
+                ];
+
+                ANDROID_HOME = "${androidSdk}/libexec/android-sdk";
+                ANDROID_NDK_HOME = androidNdk;
+                ANDROID_NDK_ROOT = androidNdk;
```
alexanderwiederin: build.rs only reads `ANDROID_NDK_HOME`, so is `ANDROID_NDK_ROOT` really needed? Worth adding a comment if it is needed.

flake.nix
```diff
@@ -85,6 +86,88 @@
...
+                                    -exec cp {} $out/lib/ \;
+                                    find target/${rustTarget}/release -path "*/out/install/include/*" \
+                                    -exec cp {} $out/include/ \;
+                  							'';
+              };
+          in
+          {
+            android-aarch64 = mkAndroidPackage "aarch64-linux-android";
+            android-armv7 = mkAndroidPackage "armv7-linux-androideabi";
+            android-x86_64 = mkAndroidPackage "x86_64-linux-android";
```
alexanderwiederin: `android-i686` is absent in line with build.rs, but not stated explicitly. If we don't want to support it, a comment like `# i686 omitted: not a current target` would close the loop.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `Block::check` to perform context-free validation of a block (size, weight, coinbase, transactions, sigops), with optional proof-of-work and merkle-root checks toggled via the `BLOCK_CHECK_BASE` / `_POW` / `_MERKLE` / `_ALL` flags. Returns a `BlockCheckResult` enum carrying the validation state on failure.
 
 ### Changed
+- Added Nix package outputs for Android with bundled NDK r27, Rust toolchains, Boost, and cmake.
```
alexanderwiederin: Should be in ### Added

README.md
```diff
@@ -34,6 +34,24 @@ dependencies. Once setup, run:
...
+
+Android cross-compilation requires [Nix](https://nixos.org/).
+
+Nix package outputs bundle the exact NDK version, Rust toolchains with Android
+targets, Boost, and cmake in a single reproducible build reduncing the support
+needed from the rust-bitcoinkernel side.
```
alexanderwiederin: I would suggest:
`Nix handles the exact NDK version, Rust toolchains, Boost, and cmake automatically, giving you a reproducible build environment with no manual setup.`

flake.nix
```diff
@@ -127,6 +138,25 @@
...
+                    "qemu-arm"
+                  else if builtins.match "x86_64.*" rustTarget != null then
+                    "qemu-x86_64"
+                  else
+                    throw "Unsupported Android target: ${rustTarget}";
+                # NDK clang wrapper triple: armv7 uses "armv7a-linux-androideabi",
+                # all others match the Rust target triple.
+                ndkClangTriple =
+                  if builtins.match "armv7.*" rustTarget != null then "armv7a-linux-androideabi" else rustTarget;
+                ndkLinker = "${androidNdk}/toolchains/llvm/prebuilt/linux-x86_64/bin/${ndkClangTriple}24-clang";
```
alexanderwiederin: `ndkLinker` hardcodes API level 24, coupling it to build.rs. Consider extracting this into a shared variable or environment variable so both can be updated in one place.

flake.nix
```diff
@@ -127,6 +138,25 @@
                   cargo = rustTargetToolchain;
                   rustc = rustTargetToolchain;
                 };
+                qemuBin =
+                  if builtins.match "aarch64.*" rustTarget != null then
+                    "qemu-aarch64"
+                  else if builtins.match "armv7.*" rustTarget != null then
+                    "qemu-arm"
+                  else if builtins.match "x86_64.*" rustTarget != null then
+                    "qemu-x86_64"
```
alexanderwiederin: Is this ever reached or is this a dead branch?

flake.nix
```diff
@@ -150,17 +183,24 @@
...
                 # derivation time, so we bypass it for cross builds.
                 dontCargoBuild = true;
-                doCheck = false;
+                doCheck = canTest;
                 buildPhase = ''
-                        					cargo build -p libbitcoinkernel-sys --target ${rustTarget} --offline --release
-                  					'';
+                  cargo build -p libbitcoinkernel-sys --target ${rustTarget} --offline --release
+                '';
+                checkPhase = pkgs.lib.optionalString canTest ''
```
alexanderwiederin, line 190: Will `checkPhase` ever execute in the CI if we only have a nix build run in `‎.github/workflows/ci.yml`?

alexanderwiederin (commented):
Also reopened https://github.com/sedited/rust-bitcoinkernel/pull/158#discussion_r3302367891 which I think was not addressed.

libbitcoinkernel-sys/build.rs
jaoleal: Exposed it as a env var

flake.nix
jaoleal: Yes, checkPhase runs wheter doCheck is true in builds. we are only excluding x86 targets

flake.nix
jaoleal: dead

jaoleal:
Done, applied @alexanderwiederin suggestions

libbitcoinkernel-sys/build.rs
```diff
@@ -2,6 +2,34 @@ use std::env;
 use std::path::Path;
 use std::process::Command;
 
+// Rust target triple -> NDK ABI name.
+fn android_abi(target: &str) -> Option<&'static str> {
```
alexanderwiederin: this can now return a `&'static str`, without the Option.

libbitcoinkernel-sys/build.rs
alexanderwiederin: How about we do the following in the outer scope:

```rust
let ndk = if is_android {
    Some(env::var("ANDROID_NDK_HOME")
        .expect("Android target detected but ANDROID_NDK_HOME is not set"))
} else {
    None
};
```

Then we can reuse it with `.as_deref().unwrap()` in the `is_android` blocks.

libbitcoinkernel-sys/build.rs
```diff
@@ -38,7 +67,45 @@ fn main() {
...
+        let ndk = env::var("ANDROID_NDK_HOME")
+            .expect("Android target detected but ANDROID_NDK_HOME is not set");
+        let toolchain_file = format!("{ndk}/build/cmake/android.toolchain.cmake");
+
+        let abi = android_abi(&target).unwrap();
+
+        // API level 24+ is required because Bitcoin Core uses getifaddrs
+        // which was introduced in Android API 24 (Nougat).
+        //
+        // This can be overriden by setting ANDROID_API_LEVEL.
```
alexanderwiederin: Typo in `overriden` -> `overridden`. I would also make it clear that this is to a higher level. 

Something like: `// This can be overridden to a higher level by setting ANDROID_API_LEVEL.`

libbitcoinkernel-sys/build.rs
```diff
@@ -70,19 +137,31 @@ fn main() {
...
     println!("cargo:rustc-link-lib=static=bitcoinkernel");
 
     let compiler = cc::Build::new().get_compiler();
-    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
 
     if target_os == "windows" {
         println!("cargo:rustc-link-lib=bcrypt");
         println!("cargo:rustc-link-lib=shell32");
-    }
+    } else if is_android {
```
alexanderwiederin, line 135: Let's also add the following here: 
```rust
println!("cargo:rerun-if-env-changed=ANDROID_NDK_HOME");
println!("cargo:rerun-if-env-changed=ANDROID_API_LEVEL");
```
Without these, changing either variable between builds won't trigger a rebuild, so you could silently end up with a stale binary built against the wrong NDK or API level.

libbitcoinkernel-sys/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Added build support for android
```
alexanderwiederin: nit: `android` -> `Android`

flake.nix
```diff
@@ -85,6 +86,89 @@
...
+                ANDROID_NDK_HOME = androidNdk;
+                CMAKE_PREFIX_PATH = "${androidPkgs.boost.dev}";
+
+                # cargoBuildHook hardcodes the host --target at
+                # derivation time, so we bypass it for cross builds.
+                dontCargoBuild = true;
+                doCheck = false;
+                buildPhase = ''
+                        					cargo build -p libbitcoinkernel-sys --target ${rustTarget} --offline --release
+                  					'';
```
alexanderwiederin: I think indentation is off here for commit 9b14b562a66d634e74bb81ae92fa83ee031a35a7

flake.nix
```diff
@@ -85,6 +86,89 @@
...
+                installPhase = ''
+                                    mkdir -p $out/lib $out/include
+                                    find target/${rustTarget}/release -path "*/out/install/lib/*.a" \
+                                    -exec cp {} $out/lib/ \;
+                                    find target/${rustTarget}/release -path "*/out/install/include/*" \
+                                    -exec cp {} $out/include/ \;
+                  							'';
```
alexanderwiederin: I also think indentation is off here for commit 9b14b562a66d634e74bb81ae92fa83ee031a35a7

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added Nix package outputs for Android with bundled NDK r24, Rust toolchains, Boost, and cmake.
```
alexanderwiederin, line 11: Should this be r27? Judging by line 93 of flake.nix.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.2.1] 2026-05-20
 
 ### Added
+- Added Nix package outputs for Android with bundled NDK r27, Rust toolchains, Boost, and cmake.
```
alexanderwiederin: This should not be here.

alexanderwiederin (commented):
Thanks for the wait!

libbitcoinkernel-sys/build.rs
jaoleal: Nice Catch! Added part of your suggestion as a comment

flake.nix
jaoleal: Nixfmt is buggy with zed, i had to remove all tabs before the whole block and manually apply `nixfmt flake.nix` to get a normal identation.

CHANGELOG.md
jaoleal: No, i had the same doubt, If you check on flake.nix, youll see that we dont  use `ANDROID_API_LEVEL = "24";` directly in the build process itself but its still declared as a variable for one to alter if needed, it would reflect also altering
 ```Rust
        // API level 24+ is required because Bitcoin Core uses getifaddrs
        // which was introduced in Android API 24 (Nougat).
        //
        // This can be overridden to a higher level by setting ANDROID_API_LEVEL.
        let api_level = match env::var("ANDROID_API_LEVEL") {
            Ok(level) => {
                let n: u32 = level.parse().expect("ANDROID_API_LEVEL must be a number");
                assert!(n >= 24, "ANDROID_API_LEVEL must be 24+");
                level
            }
            _ => "24".to_string(),
        };
```

 It appears that the API version is not the same number as the RELEASE version, the `ndkVersion = "27.2.12479018";` that we use to declare the components release. 

Ill add a comment clarifying that to avoid future confusion

**Comment by jaoleal** [comment]
Addressed @alexanderwiederin suggestions and fixes

File: CHANGELOG.md [hunk 37]
```diff
@@ -19,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.2.1] 2026-05-20
 
 ### Added
+a
```
alexanderwiederin: Can be removed in 5442965ebefc562adde3ed18c9b3bc4fe7f2177a

.github/workflows/ci.yml
```diff
@@ -218,6 +218,26 @@ jobs:
     - name: Build and test
```
alexanderwiederin: deaa42ba13a3c79efd288c68a087267bdc0aef8c commit has something wrong with the commit message

alexanderwiederin (commented):
Getting there! Could you update the PR description? I don't think it's a POC anymore. The PR also consists of four and not three commits.

CHANGELOG.md
jaoleal: Sorry, i used it as a git hack to raise a conflict and remove a thing.

jaoleal:
>Getting there!

Gosh, I entered in automatic mode for this one... Sorry

jaoleal:
Changelog:

- I reworded some commits into better and descriptive messages.
- I squashed the last commit that introduce test preparation into the commit that introduced the nix build instructions, so the nix changes are final. The commit message should also explain that testing is restricted to select capable hosts.
- I removed a variable and the newly introduced methods and inline their logic. I think this helps readers to understand whats necessary only for the android scope. Also, the less diff the better ?

jaoleal:
Just a thing, since the android outputs can only be built with nix, I still didnt found a way to integrate it on floresta upstream so what im planning is to offer the compiled binaries on floresta-nix with the current patches presented here. 

I know that this is more a floresta problem but I think other projects, those depending on rust-bitcoinkernel as floresta does, may lead into the same problem and maybe a good solution would be if this project offered its own pre-compiled Android artifacts, what you guys think about it ?

Davidson-Souza:
Since yesterday I'm experimenting with this PR and integrating Floresta with Maven, so Android devs can consume Floresta as any other library. We already had some scaffolding in [floresta-ffi](https://github.com/getfloresta/floresta-ffi) but couldn't build `bitcoinkernel` due to the things fixed here. 

I've opened https://github.com/getfloresta/floresta-ffi/pull/11 fixing it, I can now build and run floresta with kernel enabled. Writing the main pain points here for reviewers and future reference. All my experiments **did not use Nix**. I've coded this on Arch Linux and Ci runs on Ubuntu.

 - NDK couldn't find boost's headers in the host system. I had to symlink them to my ndk folder. This is kinda ugly, currently looking for a better approach
 - export `Boost_DIR` and `CMAKE_PREFIX_PATH`, which is fine
 - I also had to vendor `libc++_shared.so` because it kept loading what Android calls the `system STL`. This is a dummy `STL` with only `new/delete`. When kernel tries to use some C++ name, I would get a linker error. I'm currently digging the [docs](https://developer.android.com/ndk/guides/cpp-support) to figure a better way to do that.
 - Something during the build is using a static version of android's libc, but using the symbol at runtime without explicitly mentioning it to the linker, so it segfaults on runtime. I still didn't really get what's going on, but it seems not related to kernel or C++.

a minor nit I've found when debugging this is that sometimes `build.rs` just ignores `cmake`'s output, so it can silently fail and get a cryptic error about not finding the shared object. You can fix this with something like:

```rust
    let install_output = Command::new("cmake")
            .arg("--install")
            .arg(&build_dir)
            .arg("--config")
            .arg(build_config)
            .output()
            .unwrap();

        if !install_output.status.success() {
            let stdout = String::from_utf8_lossy(&install_output.stdout);
            let stderr = String::from_utf8_lossy(&install_output.stderr);
            panic!(
                "cmake --install failed with status {}.\nstdout:\n{}\nstderr:\n{}",
                install_output.status, stdout, stderr
            );
    }
```

sedited:
Thanks for the report @Davidson-Souza! I hope we can get rid of boost soonish, there is some work happening towards that on the Bitcoin Core side.

.github/workflows/ci.yml
```diff
@@ -218,6 +218,26 @@ jobs:
...
       run: cargo test -vv
 
+
+  android-cross-compile:
+    name: Cross-Compile for Android (${{ matrix.package }})
```
alexanderwiederin: I would suggest: `name: Android (${{ matrix.package }})`

"Cross-Compile" is redundant since that's implied by targeting Android. Also, aarch64 and armv7 actually run tests via QEMU, so "Cross-Compile" undersells what the job does

.github/workflows/ci.yml
```diff
@@ -218,6 +218,26 @@ jobs:
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - android-aarch64
+          - android-armv7
+          - android-x86_64
```
alexanderwiederin: Add a comment above `android-x86_64`:
```
# Build-only: Rust 1.71 compiler_builtins lacks f128 routines
# required by bionic on x86_64. Tests are skipped until the
# MSRV is bumped or a workaround is found. See #198 
```

.github/workflows/ci.yml
```diff
@@ -218,6 +218,26 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install Nix
+      uses: cachix/install-nix-action@v31
+
+    - name: Build
```
alexanderwiederin: Let's name this `Build and test`

If I understand correctly aarch64/armv7 run tests via QEMU

README.md
```diff
@@ -34,6 +34,23 @@ dependencies. Once setup, run:
...
+
+Android cross-compilation requires [Nix](https://nixos.org/).
+
+Nix handles the exact NDK version, Rust toolchains, Boost, and cmake
+automatically, giving you a reproducible build environment with no manual setup.
+
+```bash
+nix build .#android-aarch64
+nix build .#android-armv7
+nix build .#android-x86_64
```
alexanderwiederin: `nix build .#android-x86_64  # build only; tests skipped (see flake.nix)`

If I understand correctly.

README.md
```diff
@@ -34,6 +34,23 @@ dependencies. Once setup, run:
...
+```
+
+The resulting libraries are placed in `result/lib/`.
```
alexanderwiederin:
```suggestion
The resulting libraries and headers are placed in `result/lib/` and `result/include/`.
```

flake.nix
```diff
@@ -85,6 +86,127 @@
...
+              includeNDK = true;
+            };
+            androidSdk = androidComposition.androidsdk;
+            androidNdk = "${androidSdk}/libexec/android-sdk/ndk/${ndkVersion}";
+
+            runTests = system == "x86_64-linux";
+
+            # x86_64-linux-android is excluded from tests: Rust 1.71's
+            # compiler_builtins lacks the f128 routines (__eqtf2, __multf3, …)
+            # that bionic's static libc.a requires on x86_64.
```
alexanderwiederin, line 118: Can we add ` See #198 `? That way we make sure we are aware of this when we do bump MSRV.

alexanderwiederin (commented):
Thanks @Davidson-Souza. I've opened a PR (still in draft) for the nit in build.rs.

This looks ready to merge from my side. 

@jaoleal @Davidson-Souza, given your context on the Android integration, I'll make sure to loop you in on any future issues or PRs touching this build path.

jaoleal:
Hey folks, ill address the suggestions.

Just a thing about the last change, in order to make armv7 floresta target to work i had to patch it with

```Rust
  // The pre-compiled libcompiler_builtins for armv7-linux-androideabi
        // ships ARM EABI helper symbols tagged with @@LIBC_N (e.g.
        // __aeabi_memcpy@@LIBC_N).  When lld links a shared library or
        // executable it errors because the LIBC_N version node is not
        // defined in any version script.  --exclude-libs,ALL marks every
        // symbol pulled from static archives as local, which suppresses
        // the version-node error.
        if ndk_triple == "arm-linux-androideabi" {
            println!("cargo:rustc-link-arg=-Wl,--exclude-libs,ALL");
        }
```

jaoleal:
Rebased

jaoleal:
Okay, CHANGELOG:

- Addressed suggestions about naming and extra commenting.
- Now that davidson achieved to build the targets whitout nix, i included manual build structions that made the Android Section on README pretty verbose, i think its nice to have but tell me if you guys think its too much.
- Renamed the outputs with a libbitcoinkernel prefix to be more explicit about what they are.

jaoleal:
Consuming the nix outputs look like this https://github.com/getfloresta/floresta-nix/pull/29/changes#diff-c765aab0783c96c0d736cf34b8b34bf38166bbfa04478db4391ce45e16d6b030R40-R42

README.md
```diff
@@ -34,6 +34,101 @@ dependencies. Once setup, run:
...
+Nix handles the exact NDK version, Rust toolchains, Boost, and cmake
+automatically, giving you a reproducible build environment with no manual setup.
+
+```bash
+nix build .#libbitcoinkernel-android-aarch64
+nix build .#libbitcoinkernel-android-armv7
+nix build .#libbitcoinkernel-android-x86_64 # build only; tests skipped (see flake.nix)
+```
+
+The resulting libraries and headers are placed in `result/lib/` and `result/include.
```
alexanderwiederin: Missing the closing "\`" for `result/include`

.github/workflows/ci.yml
```diff
@@ -218,6 +218,29 @@ jobs:
...
+          # required by bionic on x86_64. Tests are skipped until the
+          # MSRV is bumped or a workaround is found. See #198 
+          - libbitcoinkernel-android-x86_64
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install Nix
+      uses: cachix/install-nix-action@v31
+
+    - name: Build and Test
```
alexanderwiederin: `- name: Build${{ matrix.package != 'libbitcoinkernel-android-x86_64' && ' and Test' || '' }}`

alexanderwiederin (commented):
Nice - thanks for doing all this! 

I had two smaller comments that build on my previous ones - they are mostly nits so feel free to ignore.

Regarding the manual setup and prebuilt `LIBBITCOINKERNEL_LIB_DIR`: Can we focus this PR only on supporting nix? The next PR could be focused on the manual setup instructions and include a CI job so we can continuously make sure it works manually. The third PR could focus on the the `LIBBITCOINKERNEL_LIB_DIR` proposal. 

I appreciate your patience here, I know this has taken a while. I think breaking these three pieces down has a lot of value!

jaoleal:
>I think breaking these three pieces down has a lot of value!

I agree with you, do you mind if i just cherry pick them out and open a pr so we can work on them later ?

alexanderwiederin:
This looks good. Thanks for pushing through!

PR #163, opened 2026-04-24
Chore: remove `bindgen` and replace with hand-rolled FFI bindings

### LINKED ISSUE #157: Consider removing `bindgen` as a build dependency
### Background

PR #150 was opened to fix a CI breakage caused by `bindgen` v0.72 pulling in `rustc-hash` v2.1.2, which introduced a rustc 1.77.0 requirement, conflicting with our MSRV of 1.71.0.

As long as `bindgen` is a build dependency, we are exposed to transitive dependency churn and MSRV breakage.

### Proposal

Maintain the FFI bindings as hand-written Rust and remove `bindgen`. Example: https://github.com/rust-bitcoin/rust-secp256k1/tree/master/secp256k1-sys

alexanderwiederin:
Closes: https://github.com/sedited/rust-bitcoinkernel/issues/157

### Summary 
Replaces the bindgen-generated FFI bindings with a hand-written `lib.rs` in `libbitcoinkernel-sys`. As a follow-up the `constants.rs` in the wrapper crate is removed and call sites are updated to use the `btck_*` names from the sys crate.

### Motivation

- Auditability
- No build-time dependency

### Changes

- `libbitcoinkernel-sys/src/lib.rs`: Hand-written FFI bindings replacing the bindgen-generated `bindings.rs`.
- `std::os::raw` type aliases replaced with `core::ffi` equivalents. 
- `libbitcoinkernel-sys/build.rs`: Bindgen block removed.
- `libbitcoinkernel-sys/Cargo.toml`: `bindgen` removed from build-dependencies.
- `src/ffi/constants.rs`: Deleted. Constants now imported from `libbitcoinkernel-sys` using `btck_*` names.
- `Cargo-minimal.lock`, `Cargo-recent.lock`: Updated to reflect removed bindgen transitive dependencies.

---
**Question to Reviewers:** Is it the right moment to do this?


alexanderwiederin:
~Will take another look later to ensure the constants have been migrated accurately.~

sedited:
> Question to Reviewers: Is it the right moment to do this?

Yes!

jaoleal:
>Question to Reviewers: Is it the right moment to do this?

Yeah, this pr will help me a lot and ill be glad to help you guys to ship it.

alexanderwiederin:
Good to hear! 

I think we are practically good to go here. One thing to note is that opaque types and grouped definitions are ordered alphabetically rather than following the C API order, as I felt it would scale easiest. The `extern "C"` declarations have the C API ordering as interdependencies require it. 

Correction: On second thought, I don't think the extern "C" declarations _need_ to be in the order of the C API either.

libbitcoinkernel-sys/src/lib.rs
```diff
@@ -1,5 +1,732 @@
...
+#[repr(C)]
+pub struct btck_ChainParameters {
+    _unused: [u8; 0],
+}
+#[repr(C)]
+pub struct btck_ChainstateManager {
+    _unused: [u8; 0],
+}
+#[repr(C)]
+pub struct btck_ChainstateManagerOptions {
```
sedited, line 145: I wonder if we should also derive Debug, Copy, and Clone for these types. Do you think there could be a downside?

libbitcoinkernel-sys/src/lib.rs
```diff
@@ -1,5 +1,732 @@
-#![allow(non_upper_case_globals)]
 #![allow(non_camel_case_types)]
```
sedited: Would be cool if we could mark it as `no_std` now.

sedited (commented):
Looks good. The only drawback I see here is that we need to be careful with copying over the correct constant values.

libbitcoinkernel-sys/src/lib.rs
alexanderwiederin: I removed them to keep the `-sys` crate minimal. `openssl-sys` takes the same approach, while `secp256k1-sys` does derive them - so there's no universal convention. For me it was mostly about keeping the surface area small.

Do you have a preference?

libbitcoinkernel-sys/src/lib.rs
```diff
@@ -1,5 +1,815 @@
...
+#![no_std]
+#![allow(non_camel_case_types, non_upper_case_globals)]
 
-include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+use core::ffi::{c_char, c_int, c_uchar, c_uint, c_void};
+
+// Primitive type aliases - alphabetical order
+
+pub type btck_BlockCheckFlags = u32;
+pub type btck_BlockValidationResult = u32;
```
jaoleal, line 14: Reading this and seeing that auditability is one of the purposes for this, perhaps we could work out a little more docs here ? A good example that I saw was [secp256k1 bindings](https://github.com/rust-bitcoin/rust-secp256k1/blob/9fc86f756f0cd24a5167c131551b31b748c8cdb3/secp256k1-sys/src/lib.rs#L82).

For this case i suggest some simple docs such as
```suggestion
/// Represents a granular reason for block validation.
pub type btck_BlockValidationResult = u32;
```

libbitcoinkernel-sys/src/lib.rs
```diff
@@ -16,6 +16,88 @@ pub type btck_SynchronizationState = u8;
 pub type btck_ValidationMode = u8;
 pub type btck_Warning = u8;
 
+// btck_BlockValidationResult
```
jaoleal, line 24: The constants looks fine, cant we maintain the `constants.rs` ? That would leave this lib.rs only for declarations right ?

jaoleal (approved):
ACK 99759099cd305e487e0d0a9438e3cd843d0e16d5

Only nits and docs suggestions, perhaps the second commit thats solely to move constants could be included in CHANGELOG.md ?

Also, the README.md under `libbitcoinkernel-sys` is pretty simple, we could include some sayings about the hand written bindings

jaoleal:
Question: will this trigger a new release ?

sedited:
Yes :)

libbitcoinkernel-sys/src/lib.rs
```diff
@@ -1,5 +1,815 @@
...
+}
+
+// Compile-time layout guards - catch any ABI drift
+const _: () = {
+    assert!(core::mem::size_of::<btck_LoggingOptions>() == 20);
+    assert!(core::mem::align_of::<btck_LoggingOptions>() == 4);
+    assert!(core::mem::size_of::<btck_NotificationInterfaceCallbacks>() == 72);
+    assert!(core::mem::align_of::<btck_NotificationInterfaceCallbacks>() == 8);
+    assert!(core::mem::size_of::<btck_ValidationInterfaceCallbacks>() == 48);
+    assert!(core::mem::align_of::<btck_ValidationInterfaceCallbacks>() == 8);
```
jaoleal:
```suggestion
   assert!(
        core::mem::size_of::<btck_NotificationInterfaceCallbacks>()
            == 9 * core::mem::size_of::<*const ()>()
    );
    assert!(
        core::mem::align_of::<btck_NotificationInterfaceCallbacks>()
            == core::mem::align_of::<*const ()>()
    );
    assert!(
        core::mem::size_of::<btck_ValidationInterfaceCallbacks>()
            == 6 * core::mem::size_of::<*const ()>()
    );
    assert!(
        core::mem::align_of::<btck_ValidationInterfaceCallbacks>()
            == core::mem::align_of::<*const ()>()
    );
```


After rebasing #158 on top of this to see if anything breaks, i found a problem regarding pointers size on `armv7-linux-androideabi` since pointers are 4 bytes on 32-bit targets.

the above suggestion fixed the problem and i was able to successfully build this lib to `armv7-linux-androideabi`. It multiplies the size of pointers at compile time but im not totally sure if its the best fix for this case.

alexanderwiederin: Good catch! The hardcoded values were assuming 64-but pointer sizes. Your suggestion make the assertions valid on both 32-bit and 64-bit targets.

Applied - thanks!

libbitcoinkernel-sys/src/lib.rs
alexanderwiederin: I would prefer to keep everything in `lib.rs` and mirror the structure of the C header. This makes auditing easier, especially when updating to newer versions. The file is organised with the section comments to make navigation straightforward. Having said that, if the API grows significantly we can revisit splitting it up.

libbitcoinkernel-sys/src/lib.rs
alexanderwiederin: Good suggestion, but I'd like to keep the scope focused on the bindings as much as possible. I'll add a top-level doc comment and update the README pointing to the C header for now.

alexanderwiederin:
> Only nits and docs suggestions, perhaps the second commit thats solely to move constants could be included in CHANGELOG.md ?

Done.

> Also, the README.md under `libbitcoinkernel-sys` is pretty simple, we could include some sayings about the hand written bindings

Done.

Thanks for the feedback!

libbitcoinkernel-sys/src/lib.rs
alexanderwiederin: Added https://github.com/sedited/rust-bitcoinkernel/issues/175 to formally support 32-bit targets.

jaoleal (approved):
ACK

jaoleal:
ACK ccb3542

jaoleal:
Anything holding this ?

alexanderwiederin:
> Anything holding this ?

Not from my side. Thanks for bumping.

sedited:
Sorry for the wait here. This looks fine to me, maybe some of the struct alignment checks are a bit brittle, but we can fix them up if they ever break. 

ACK

Thank you @alexanderwiederin and the reviews @jaoleal!

PR #167, opened 2026-04-28
fix(core): use &self in BlockTreeEntry::prev for consistency

alexanderwiederin:
### Motivation

All other methods take `&self`. `BlockTreeEntry`s are `Copy`, so the values are never actually consumed, but are just copied.


sedited (approved):
Seems like this was left over from a long time ago. LGTM

PR #170, opened 2026-04-29
chore(ci): add lock file check job

alexanderwiederin:
### Summary

Adds a CI job to check the minimal lock file and detect drift. Only cargo-minimal.lock is checked since Cargo-recent.lock drifts with every new crate release, making it too noisy to enforce in CI.

### Motivation

Removes the need to run the commands locally to verify changes are correct.


sedited (approved):
LGTM

PR #176, opened 2026-05-07
chore(ci): add nix flake check

alexanderwiederin:
### Summary

Adds a check to the CI to ensure the Nix flake works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,25 @@ jobs:
         cargo +nightly update -Z minimal-versions
         git diff --exit-code Cargo-minimal.lock
 
+  nix-flake-check:
+    name: Check Nix flake
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


jaoleal:
This is nice, have you considered to add statix and nixfmt to checks too ? They can help to maintain the flake well writen. Heres an example on a recent change of mine.

https://github.com/getfloresta/Floresta/blob/22fa75a929f95d287806deed96f56619f36e7459/flake.nix#L28

jaoleal:
Actually, I ask for statix and nixfmt to be added since this flake appears to be strangely formatted and my editor keeps reformating it all the time that I forgot to save without formatting... 😭

alexanderwiederin:
Good call on nixfmt! I think it would be best if we merge this first and then edit the flake with the formatting changes - that way we are sure there are no functional changes.

jaoleal (approved):
ACK 40e4db3c53444eba2b8f2ccd86d075b9798c0ca8

PR #177, opened 2026-05-07
feat(core): expose Block::check

### LINKED ISSUE #155: Add context-free Block Validation
Implement context-free Block Validation in the API.

https://github.com/bitcoin/bitcoin/pull/33908

pzafonte:
Wraps btck_block_check from bitcoin/bitcoin#33908. The function performs, context-free validation of a block: size, weight, coinbase, transactions, sigops, without chainstate or block index access. Proof-of-work and merkle-root checks are optional via the BLOCK_CHECK_BASE / _POW / _MERKLE / _ALL flags.

Exposed as Block::check returning BlockCheckResult { Valid, Invalid(BlockValidationState) }. Raw btck_BlockCheckFlags_* constants are added to the sys crate's hand-rolled bindings; user-facing BLOCK_CHECK_* aliases in src/core/block.rs are re-exported via a block_check_flags submodule.

Tests are split per scenario as unit tests in src/core/block.rs (valid block, mutated merkle root, invalid PoW, tampered coinbase). A follow-up commit reuses MAINNET_BLOCK_1_HEX in test_block_hash_display and test_block_hash_ref_display since they validate the same block.

Closes #155.

### DIFF
--- a/src/core/block.rs
+++ b/src/core/block.rs
@@ -128,31 +128,60 @@ use std::{
 };
 
 use libbitcoinkernel_sys::{
-    btck_Block, btck_BlockHash, btck_BlockHeader, btck_BlockSpentOutputs, btck_Coin,
-    btck_TransactionSpentOutputs, btck_block_copy, btck_block_count_transactions,
-    btck_block_create, btck_block_destroy, btck_block_get_hash, btck_block_get_header,
-    btck_block_get_transaction_at, btck_block_hash_copy, btck_block_hash_create,
-    btck_block_hash_destroy, btck_block_hash_equals, btck_block_hash_to_bytes,
-    btck_block_header_copy, btck_block_header_create, btck_block_header_destroy,
-    btck_block_header_get_bits, btck_block_header_get_hash, btck_block_header_get_nonce,
-    btck_block_header_get_prev_hash, btck_block_header_get_timestamp,
+    btck_Block, btck_BlockCheckFlags, btck_BlockCheckFlags_ALL, btck_BlockCheckFlags_BASE,
+    btck_BlockCheckFlags_MERKLE, btck_BlockCheckFlags_POW, btck_BlockHash, btck_BlockHeader,
+    btck_BlockSpentOutputs, btck_Coin, btck_TransactionSpentOutputs, btck_block_check,
+    btck_block_copy, btck_block_count_transactions, btck_block_create, btck_block_destroy,
+    btck_block_get_hash, btck_block_get_header, btck_block_get_transaction_at,
+    btck_block_hash_copy, btck_block_hash_create, btck_block_hash_destroy, btck_block_hash_equals,
+    btck_block_hash_to_bytes, btck_block_header_copy, btck_block_header_create,
+    btck_block_header_destroy, btck_block_header_get_bits, btck_block_header_get_hash,
+    btck_block_header_get_nonce, btck_block_header_get_prev_hash, btck_block_header_get_timestamp,
     btck_block_header_get_version, btck_block_header_to_bytes, btck_block_spent_outputs_copy,
     btck_block_spent_outputs_count, btck_block_spent_outputs_destroy,
     btck_block_spent_outputs_get_transaction_spent_outputs_at, btck_block_to_bytes,
-    btck_coin_confirmation_height, btck_coin_copy, btck_coin_destroy, btck_coin_get_output,
-    btck_coin_is_coinbase, btck_transaction_spent_outputs_copy,
-    btck_transaction_spent_outputs_count, btck_transaction_spent_outputs_destroy,
-    btck_transaction_spent_outputs_get_coin_at,
+    btck_chain_parameters_get_consensus_params, btck_coin_confirmation_height, btck_coin_copy,
+    btck_coin_destroy, btck_coin_get_output, btck_coin_is_coinbase,
+    btck_transaction_spent_outputs_copy, btck_transaction_spent_outputs_count,
+    btck_transaction_spent_outputs_destroy, btck_transaction_spent_outputs_get_coin_at,
 };
 
 use crate::{
     c_helpers, c_serialize,
     ffi::{
         c_helpers::present,
-        sealed::{AsPtr, FromMutPtr, FromPtr},
+        sealed::{AsMutPtr, AsPtr, FromMutPtr, FromPtr},
     },
+    notifications::types::BlockValidationState,
+    state::context::ChainParams,
     KernelError,
 };
+
+/// Bitmask of flags controlling which checks [`Block::check`] performs.
+pub type BlockCheckFlags = btck_BlockCheckFlags;
+
+/// Run only base context-free block checks (no PoW, no Merkle root).
+pub const BLOCK_CHECK_BASE: BlockCheckFlags = btck_BlockCheckFlags_BASE;
+
+/// Enable Proof-of-Work verification via the block header.
+pub const BLOCK_CHECK_POW: BlockCheckFlags = btck_BlockCheckFlags_POW;
+
+/// Enable Merkle-root verification (and mutation detection).
+pub const BLOCK_CHECK_MERKLE: BlockCheckFlags = btck_BlockCheckFlags_MERKLE;
+
+/// Enable all available context-free block checks (PoW + Merkle root).
+pub const BLOCK_CHECK_ALL: BlockCheckFlags = btck_BlockCheckFlags_ALL;
+
+/// Outcome of [`Block::check`].
+///
+/// On failure, the [`BlockValidationState`] carries details that can be
+/// inspected via [`BlockValidationStateExt`](crate::notifications::BlockValidationStateExt).
+pub enum BlockCheckResult {
+    /// The block passed the requested context-free checks.
+    Valid,
+    /// The block failed; the state holds the validation details.
+    Invalid(BlockValidationState),
+}
 
 use super::transaction::{TransactionRef, TxOutRef};
 
@@ -941,6 +970,51 @@ impl Block {
     /// ```
     pub fn transactions(&self) -> BlockTransactionIter<'_> {
         BlockTransactionIter::new(self)
+    }
+
+    /// Performs context-free validation checks on this block.
+    ///
+    /// Runs base structural checks (size, weight, coinbase, transactions,
+    /// sigops) without requiring chainstate or block index access.
+    /// Proof-of-work and merkle-root checks are optional and toggled via `flags`.
+    ///
+    /// # Arguments
+    /// * `chain_params` - Chain parameters providing consensus rules
+    /// * `flags` - Bitmask of [`BLOCK_CHECK_BASE`], [`BLOCK_CHECK_POW`],
+    ///   [`BLOCK_CHECK_MERKLE`], or [`BLOCK_CHECK_ALL`]
+    ///
+    /// # Returns
+    /// [`BlockCheckResult::Valid`] on success, otherwise
+    /// [`BlockCheckResult::Invalid`] carrying the validation state.
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # use bitcoinkernel::{
+    /// #     prelude::*, Block, BlockCheckResult, ChainParams, ChainType, BLOCK_CHECK_ALL,
+    /// # };
+    /// # fn example() -> Result<(), bitcoinkernel::KernelError> {
+    /// # let block_data = vec![0u8; 100]; // placeholder
+    /// # let block = Block::new(&block_data)?;
+    /// let chain_params = ChainParams::new(ChainType::Mainnet);
+    ///
+    /// match block.check(&chain_params, BLOCK_CHECK_ALL) {
+    ///     BlockCheckResult::Valid => println!("ok"),
+    ///     BlockCheckResult::Invalid(state) => println!("failed: {:?}", state.result()),
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn check(&self, chain_params: &ChainParams, flags: BlockCheckFlags) -> BlockCheckResult {
+        let state = BlockValidationState::new();
+        let consensus_params =
+            unsafe { btck_chain_parameters_get_consensus_params(chain_params.as_ptr()) };
+        let result =
+            unsafe { btck_block_check(self.inner, consensus_params, flags, state.as_mut_ptr()) };
+        if c_helpers::verification_passed(result) {
+            BlockCheckResult::Valid
+        } else {
+            BlockCheckResult::Invalid(state)
+        }
     }
 }
 
@@ -1865,6 +1939,8 @@ mod tests {
     use crate::ffi::test_utils::{
         test_owned_clone_and_send, test_owned_trait_requirements, test_ref_trait_requirements,
     };
+    use crate::prelude::*;
+    use crate::{BlockValidationResult, ChainType, ValidationMode};
     use std::{
         fs::File,
         io::{BufRead, BufReader},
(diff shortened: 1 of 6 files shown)


sedited:
Can you rebase and undraft this?

src/core/block.rs
```diff
@@ -942,6 +956,57 @@ impl Block {
...
+    /// let chain_params = ChainParams::new(ChainType::Mainnet);
+    /// let mut state = BlockValidationState::new();
+    ///
+    /// if !block.check_context_free(&chain_params, BLOCK_CHECK_ALL, &mut state) {
+    ///     println!("Failed: {:?}", state.result());
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn check_context_free(
```
alexanderwiederin: I think we can call the method `check()`. What do you think?

src/core/block.rs
```diff
@@ -942,6 +956,57 @@ impl Block {
...
+        &self,
+        chain_params: &ChainParams,
+        flags: btck_BlockCheckFlags,
+        validation_state: &mut BlockValidationState,
```
alexanderwiederin: Let's create the state in the method and return a new one every time. i.e. remove `validation_state` from the function params and create it in the method. I think it's better experience. See `ChainstateManager::process_block_header` for reference.

src/core/block.rs
```diff
@@ -942,6 +956,57 @@ impl Block {
...
+    ) -> bool {
```
alexanderwiederin: I would suggest something like this: 

```
pub enum BlockCheckResult {
    Valid,
    Invalid(BlockValidationState),
}
```

Instead of a bool return value. 

See `ChainstateManager::process_block_header()`

src/core/block.rs
```diff
@@ -151,9 +151,23 @@ use crate::{
...
+
+/// Enable Proof-of-Work verification via the block header.
+pub const BLOCK_CHECK_POW: btck_BlockCheckFlags = 1 << 0;
+
+/// Enable Merkle-root verification (and mutation detection).
+pub const BLOCK_CHECK_MERKLE: btck_BlockCheckFlags = 1 << 1;
+
+/// Enable all available context-free block checks (PoW + Merkle root).
+pub const BLOCK_CHECK_ALL: btck_BlockCheckFlags = BLOCK_CHECK_POW | BLOCK_CHECK_MERKLE;
+
```
alexanderwiederin, line 185: These should go in constants.rs

src/core/block.rs
```diff
@@ -942,6 +956,57 @@ impl Block {
...
+        validation_state: &mut BlockValidationState,
+    ) -> bool {
+        let consensus_params =
+            unsafe { btck_chain_parameters_get_consensus_params(chain_params.as_ptr()) };
+        let result = unsafe {
+            btck_block_check(
+                self.inner,
+                consensus_params,
+                flags,
+                validation_state.as_ptr() as *mut btck_BlockValidationState,
```
alexanderwiederin: The `.as_ptr() as *mut` is a bit of a faux pas. Added [PR](https://github.com/sedited/rust-bitcoinkernel/pull/178) to remove the implicit cast, which can be used here.

tests/test.rs
```diff
@@ -918,4 +919,78 @@ mod tests {
         // is_sync::<Rc<u8>>(); // won't compile, kept as a failure case.
         // is_send::<Rc<u8>>(); // won't compile, kept as a failure case.
     }
+
+    // Mainnet block 1.
```
alexanderwiederin: This comment can be removed. The variable name is already explicit enough.

tests/test.rs
```diff
@@ -918,4 +919,78 @@ mod tests {
...
+    const MAINNET_BLOCK_1_HEX: &str = "010000006fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c\
+        68d6190000000000982051fd1e4ba744bbbe680e1fee14677ba1a3c3540bf7b1cdb606e857233e0e61bc6649ff\
+        ff001d01e362990101000000010000000000000000000000000000000000000000000000000000000000000000\
+        ffffffff0704ffff001d0104ffffffff0100f2052a0100000043410496b538e853519c726a2c91e61ec11600ae\
+        1390813a627c66fb8be7947be63c52da7589379515d4e0a604f8141781e62294721166bf621e73a82cbf2342c8\
+        58eeac00000000";
+
+    #[test]
+    fn test_block_check_context_free() {
```
alexanderwiederin: I think these tests can live in `core/block.rs` and be broken down into smaller individual tests.`tests/test.rs` is _ideally_ focused on integration tests.

tests/test.rs
```diff
@@ -918,4 +919,78 @@ mod tests {
...
+        // Byte offsets within the serialized block.
+        const MERKLE_ROOT_OFFSET: usize = 4 + 32;
+        const NBITS_OFFSET: usize = 4 + 32 + 32 + 4;
+        const COINBASE_PREVOUT_N_OFFSET: usize = 4 + 32 + 32 + 4 + 4 + 4 + 1 + 4 + 1 + 32;
```
alexanderwiederin: It would be great if we had the labels of the offsets like this:
```
const COINBASE_PREVOUT_N_OFFSET: usize =
        4     // version
        + 32  // prev_hash
        + 32  // merkle_root
        + 4   // timestamp
        + 4   // bits
        + 4   // nonce
        + 1   // tx count varint
        + 4   // tx version
        + 1   // vin count
        + 32; // prevout hash
```

tests/test.rs
```diff
@@ -918,4 +919,78 @@ mod tests {
...
+        let chain_params = ChainParams::new(ChainType::Mainnet);
+
+        let block = Block::new(&raw_block).unwrap();
+        let mut state = BlockValidationState::new();
+
+        assert!(block.check_context_free(&chain_params, BLOCK_CHECK_BASE, &mut state));
+        assert_eq!(state.mode(), ValidationMode::Valid);
+
+        assert!(block.check_context_free(&chain_params, BLOCK_CHECK_ALL, &mut state));
+        assert_eq!(state.mode(), ValidationMode::Valid);
```
alexanderwiederin: Can we also have a `BLOCK_CHECK_ALL` test for an invalid block?

tests/test.rs
```diff
@@ -918,4 +919,78 @@ mod tests {
...
+        let truncated_block_data = hex::decode(
+            "010000006fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000\
+             982051fd1e4ba744bbbe680e1fee14677ba1a3c3540bf7b1cdb606e857233e0e61bc6649\
+             ffff001d01e36299",
+        )
+        .unwrap();
+        assert!(matches!(
+            Block::new(&truncated_block_data),
+            Err(KernelError::Internal(_))
+        ));
```
alexanderwiederin: This case is already covered in `core/block.rs`

alexanderwiederin (changes requested):
Cool! 

Could you also please add the change to the `CHANGELOG`?

pzafonte:
Thanks for the review. I addressed everything in the latest commit except the `as_ptr() as *mut`. I'll rebase and switch the cast in `Block::check` to `as_mut_ptr()` once #178 goes through.

src/core/block.rs
```diff
@@ -2288,4 +2370,127 @@ mod tests {
...
+
+        // Mutating the merkle root also alters the block hash, so the
+        // combined ALL check fails on the PoW path before the merkle
+        // path runs.
+        match block.check(&chain_params, BLOCK_CHECK_ALL) {
+            BlockCheckResult::Invalid(state) => {
+                assert_eq!(state.mode(), ValidationMode::Invalid);
+                assert_eq!(state.result(), BlockValidationResult::InvalidHeader);
+            }
+            _ => assert!(false),
```
alexanderwiederin: I would suggest to remove the `assert!(false)` for this line and the other occurences with: 
```suggestion
            _ => unreachable!("expected BlockCheckResult::Invalid"),
```

src/core/block.rs
```diff
@@ -128,32 +128,59 @@ use std::{
...
+pub const BLOCK_CHECK_MERKLE: btck_BlockCheckFlags = BTCK_BLOCK_CHECK_FLAGS_MERKLE;
+
+/// Enable all available context-free block checks (PoW + Merkle root).
+pub const BLOCK_CHECK_ALL: btck_BlockCheckFlags = BTCK_BLOCK_CHECK_FLAGS_ALL;
+
+/// Outcome of [`Block::check`].
+///
+/// On failure, the [`BlockValidationState`] carries details that can be
+/// inspected via [`BlockValidationStateExt`](crate::notifications::BlockValidationStateExt).
+pub enum BlockCheckResult {
```
alexanderwiederin, line 179: `BlockCheckResult` should derive `Debug` once `BlockValidationState` implements it. Added https://github.com/sedited/rust-bitcoinkernel/issues/181 for this. Nothing to do or add here for now.

src/core/block.rs
```diff
@@ -2288,4 +2370,127 @@ mod tests {
             "00000000839a8e6886ab5951d76f411475428afc90947ee320161bbf18eb6048"
         );
     }
+
+    const MAINNET_BLOCK_1_HEX: &str = "010000006fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c\
+        68d6190000000000982051fd1e4ba744bbbe680e1fee14677ba1a3c3540bf7b1cdb606e857233e0e61bc6649ff\
+        ff001d01e362990101000000010000000000000000000000000000000000000000000000000000000000000000\
+        ffffffff0704ffff001d0104ffffffff0100f2052a0100000043410496b538e853519c726a2c91e61ec11600ae\
+        1390813a627c66fb8be7947be63c52da7589379515d4e0a604f8141781e62294721166bf621e73a82cbf2342c8\
+        58eeac00000000";
```
alexanderwiederin: This is the same block as used in `test_block_hash_display` and `test_block_hash_ref_display`. Can we set the variable at the top of the test module? If you have time, we could add another commit to use that variable in those two tests as well, if not happy to take that on.

alexanderwiederin (commented):
Nice turnaround.

Please rebase for #178 and #179.

The question on the method name was genuine; I am not sure whether `Block::check()` is better than `Block::check_context_free()`. Any thoughts from you or other reviewers?

pzafonte:
> The question on the method name was genuine; I am not sure whether `Block::check()` is better than `Block::check_context_free()`. Any thoughts from you or other reviewers?

I lean towards `Block::check_context_free()` because it's self-documenting and there could plausibly be a contextual block check down the road.

pzafonte:
Changes updated in 26121eb

Reuses `MAINNET_BLOCK_1_HEX` in `test_block_hash_display` and `test_block_hash_ref_display` in 60e9bf6

src/core/block.rs
```diff
@@ -2288,4 +2347,120 @@ mod tests {
             "00000000839a8e6886ab5951d76f411475428afc90947ee320161bbf18eb6048"
         );
     }
+
+    #[test]
+    fn check_valid_block_passes_base_and_all() {
+        use crate::prelude::*;
+        use crate::ChainType;
```
alexanderwiederin: Lets move these to the top of the test module to the other imports. Same goes for the occurrences below.

alexanderwiederin (commented):
> I lean towards Block::check_context_free() because it's self-documenting and there could plausibly be a contextual block check down the road.

Thanks for this. I do think there is merit to the `check_context_free`, but I doubt that a contextual block check would be called "check". The current proposals lean towards "connect" (https://github.com/bitcoin/bitcoin/pull/32317) or "validate" (https://github.com/bitcoin/bitcoin/pull/35187) and would be on the chainstate manager. 

Either way, I wanted to hear your opinion on this.

I think check is more idiomatic Rust - discoverability is better addressed by the doc comments you've already added and the flags parameter, which signal what kind of check this is at the call site.

sedited:
@pzafonte this will require a rebase and probably a change to the bindings now.

pzafonte:
Makes sense, if other contextual variants use  `connect` or `validate` then there's no collision. If any future `check` arrives then that method can take the qualifier, which is why I implicitly agreed with your suggestion.

Rebased on current master. Addressed the remaining feedback.

src/core/block.rs
```diff
@@ -128,32 +128,58 @@ use std::{
...
+pub const BLOCK_CHECK_BASE: btck_BlockCheckFlags = btck_BlockCheckFlags_BASE;
+
+/// Enable Proof-of-Work verification via the block header.
+pub const BLOCK_CHECK_POW: btck_BlockCheckFlags = btck_BlockCheckFlags_POW;
+
+/// Enable Merkle-root verification (and mutation detection).
+pub const BLOCK_CHECK_MERKLE: btck_BlockCheckFlags = btck_BlockCheckFlags_MERKLE;
+
+/// Enable all available context-free block checks (PoW + Merkle root).
+pub const BLOCK_CHECK_ALL: btck_BlockCheckFlags = btck_BlockCheckFlags_ALL;
```
alexanderwiederin:
```suggestion
pub type BlockCheckFlags = btck_BlockCheckFlags;

/// Run only base context-free block checks (no PoW, no Merkle root).
pub const BLOCK_CHECK_BASE: BlockCheckFlags = btck_BlockCheckFlags_BASE;
/// Enable Proof-of-Work verification via the block header.
pub const BLOCK_CHECK_POW: BlockCheckFlags = btck_BlockCheckFlags_POW;
/// Enable Merkle-root verification (and mutation detection).
pub const BLOCK_CHECK_MERKLE: BlockCheckFlags = btck_BlockCheckFlags_MERKLE;
/// Enable all available context-free block checks (PoW + Merkle root).
pub const BLOCK_CHECK_ALL: BlockCheckFlags = btck_BlockCheckFlags_ALL;
```

Note that I added `BlockCheckFlags` to remove any mention of `btck_*` in the public API.

src/core/block.rs
```diff
@@ -942,6 +968,55 @@ impl Block {
...
+    ///     BlockCheckResult::Valid => println!("ok"),
+    ///     BlockCheckResult::Invalid(state) => println!("failed: {:?}", state.result()),
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn check(
+        &self,
+        chain_params: &ChainParams,
+        flags: btck_BlockCheckFlags,
```
alexanderwiederin:
```suggestion
        flags: BlockCheckFlags,
```
So we can avoid exposing any mention of `btck_BlockCheckFlags` in the public API.

alexanderwiederin: Opened https://github.com/sedited/rust-bitcoinkernel/pull/185 to align `verify`

src/core/block.rs
```diff
@@ -157,17 +157,19 @@ use crate::{
...
-pub const BLOCK_CHECK_POW: btck_BlockCheckFlags = btck_BlockCheckFlags_POW;
+pub const BLOCK_CHECK_POW: BlockCheckFlags = btck_BlockCheckFlags_POW;
 
 /// Enable Merkle-root verification (and mutation detection).
-pub const BLOCK_CHECK_MERKLE: btck_BlockCheckFlags = btck_BlockCheckFlags_MERKLE;
+pub const BLOCK_CHECK_MERKLE: BlockCheckFlags = btck_BlockCheckFlags_MERKLE;
 
 /// Enable all available context-free block checks (PoW + Merkle root).
-pub const BLOCK_CHECK_ALL: btck_BlockCheckFlags = btck_BlockCheckFlags_ALL;
+pub const BLOCK_CHECK_ALL: BlockCheckFlags = btck_BlockCheckFlags_ALL;
```
alexanderwiederin, line 173: This should be in the previous commit.

alexanderwiederin (commented):
Almost there - thanks for pushing through!

src/core/block.rs
```diff
@@ -128,32 +128,60 @@ use std::{
...
     c_helpers, c_serialize,
     ffi::{
         c_helpers::present,
-        sealed::{AsPtr, FromMutPtr, FromPtr},
+        sealed::{AsMutPtr, AsPtr, FromMutPtr, FromPtr},
     },
+    notifications::types::BlockValidationState,
+    state::context::ChainParams,
     KernelError,
 };
```
alexanderwiederin, line 159:
```suggestion
/// Bitmask of flags controlling which checks [`Block::check`] performs.
```
I think a doc comment would make sense here.

alexanderwiederin (commented):
Please resolve the conflicts and adhere to body line length in the second commit message according to https://cbea.ms/git-commit/ and we are good to go.

alexanderwiederin (approved):
Looks good. Thank you!

PR #178, opened 2026-05-15
refactor(ffi): add `AsMutPtr` trait and replace implicit const-to-mut pointer cast

alexanderwiederin:
### Changes

- Introduce AsMutPtr<T> to the sealed FFI trait module to properly express mutable pointer access. 
- Implement AsMutPtr for BlockValidationState (owned, *mut) but not BlockValidationStateRef (borrowed, *const)
- Replaces the `as_ptr() as *mut` cast in process_block_header with the `as_mut_ptr()` call

### Motivation

The `as_ptr() as *mut` cast worked because BlockValidationState owns a *mut internally, making it a safe roundtrip. Leaving this pattern in the codebase sets a precedent that could be copied onto types which hold a genuine *const, where casting to *mut and mutating through it would be UB.

### DIFF
--- a/src/ffi/mod.rs
+++ b/src/ffi/mod.rs
@@ -9,6 +9,11 @@ pub(crate) mod sealed {
         fn as_ptr(&self) -> *const T;
     }
 
+    pub trait AsMutPtr<T> {
+        /// Returns a raw mutable pointer to the underlying C object.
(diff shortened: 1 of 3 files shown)


sedited (approved):
Nice, LGTM

PR #184, opened 2026-05-21
fix(docs): align `BlockValidationResult` doc comments with C API

alexanderwiederin:
Several doc comments were inaccurate or vaguer than the upstream definitions. Importantly, `Consensus` was documented as "Block is valid according to consensus rules" when it actually means the opposite - a consensus rule violation. Align all affected variants with the upstream comments for accuracy.


sedited:
LGTM Thanks for patching this, I hope this is the last instance of my old, sloppy docs.

PR #185, opened 2026-05-21
fix(core): replace btck_ScriptVerificationFlags with a public type alias

alexanderwiederin:
btck_ScriptVerificationFlags was exposed in the public API via the types on the VERIFY_* constants. Add a ScriptVerificationFlags type alias and update the constants to use it.

### DIFF
--- a/src/core/verify.rs
+++ b/src/core/verify.rs
@@ -171,37 +171,40 @@ use crate::{
     c_helpers, ffi::sealed::AsPtr, KernelError, ScriptPubkeyExt, TransactionExt, TxOutExt,
 };
 
+/// Bitmask of flags controlling which consensus rules [`verify`] enforces.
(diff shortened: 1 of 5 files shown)


src/core/verify.rs
```diff
@@ -171,37 +171,39 @@ use crate::{
     c_helpers, ffi::sealed::AsPtr, KernelError, ScriptPubkeyExt, TransactionExt, TxOutExt,
 };
 
+pub type ScriptVerificationFlags = btck_ScriptVerificationFlags;
```
sedited, line 175: Can we use this in the tests too isntead of its btck_ name? I think that might take an additional export for this module.

alexanderwiederin: Done - thank you! I also added a doc comment for `ScriptVerificationFlags`

sedited (approved):
LGTM

PR #186, opened 2026-05-21
feat(core): add zero-copy `Deref` bridge to `bitcoin::Script`

### LINKED ISSUE #20: Feat: Script Type Detection
## Feature Request: Script Type Detection

Add script type detection methods to identify common Bitcoin script patterns (Taproot, OP_RETURN, P2SH, etc.).

### Proposed API
```rust
pub trait ScriptPubkeyExt {
   fn is_taproot(&self) -> bool;
   fn is_op_return(&self) -> bool;
   fn is_p2sh(&self) -> bool;
}
```

*Note: should probably expose bitcoin core's existing script classification functions through the C API (.e.g `IsPayToScriptHash`, `IsPayToWitnessScriptHash`, etc.) but is out of scope for the current state of the project*

alexanderwiederin:
closes https://github.com/sedited/rust-bitcoinkernel/issues/20

## Summary
Implements `Deref<Target = bitcoin::Script>` for `ScriptPubkey` and `ScriptPubkeyRef`, behind a new `bitcoin` feature flag. The `Deref` implementation uses a new function, `script_as_bytes`, which captures the raw pointer passed to the `btck_script_pubkey_to_bytes` callback and casts it to a `&[u8]` slice, avoiding any allocation. 

## Motivation

Silent payments block scanning requires script introspection such as `is_p2tr` ([reference](https://github.com/kernel-node/kernel-node/blob/master/crates/wallet/src/silentpayments/scanning.rs#L105)). Delegating this functionality to `rust-bitcoin` keeps the kernel C API focused on consensus, avoiding scope creep. Further, integrating `ScriptPubkey` with rust-bitcoin's `Script` type, provides a suite of new functionality with Bitcoin protocol awareness.

## Usage

```rust
use bitcoinkernel::ScriptPubkey;

let bytes = hex::decode("5120deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef").unwrap();
let script = ScriptPubkey::new(&bytes).unwrap();

// the following functionality is exposed through `bitcoin::Script`
assert!(script.is_p2tr());
assert!(script.witness_version().is_some());
assert!(!script.is_op_return());
println!("{}", script.to_asm_string());
```

## What you should know
Given that `Deref` is intended to be used for wrappers and their underlying type, the use here may be considered misplaced - `bitcoin::Script` is not an underlying type, but a _peer_ type from a separate crate that shares the same underlying representation. The alternative would be to add a `ScriptPubkey::as_bitcoin_script()` method to make the relationship explicit, at the cost of call-site ergonomics. I recognise this is a design decision that requires careful consideration. I'm happy to go either way. 

See the Rust [documentation](https://doc.rust-lang.org/std/ops/trait.Deref.html#when-to-implement-deref-or-derefmut) on when to implement `Deref`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,9 @@ jobs:
     - name: Run tests
       run: cargo test -vv
 
+    - name: Run tests (bitcoin feature)
+      run: cargo test -vv --features bitcoin
+
   linux-aarch64:
     name: Build and Test on Linux ARM64
     runs-on: ubuntu-24.04-arm
@@ -198,6 +201,9 @@ jobs:
     - name: Build and test
       run: cargo test -vv
 
+    - name: Run tests (bitcoin feature)
+      run: cargo test -vv --features bitcoin
+
   macos-build:
     name: Build and Test on macOS
     runs-on: macos-latest
@@ -217,6 +223,9 @@ jobs:
 
     - name: Build and test
       run: cargo test -vv
+
+    - name: Run tests (bitcoin feature)
+      run: cargo test -vv --features bitcoin
 
   fuzz-corpus:
     name: Verify Fuzz Corpus
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,13 +39,19 @@ jobs:
 
     - name: Build with MSRV
       run: |
-        # Ensure we're using the right version
         rustc --version
-        # Build the project
-        cargo build --all-features --locked
+        cargo build --locked
+
+    - name: Build with MSRV (bitcoin feature)
+      run: |
+        rustc --version
+        cargo build --features bitcoin --locked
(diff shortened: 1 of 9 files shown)


rustaceanrob:
Looks correct however it is a shame we won't be able to do this deref trick for other types in the `bitcoin` crate. Another improvement on the current situation would be to allow users to call something like `as_bytes -> &[u8]`. This would allow some analysis of the script (i.e. is it a script type I care about) without allocating the vector with `to_bytes`. Does that seem plausible to add without the addition of `bitcoin`? There is a function that does something similar that was introduced internally as part of this PR.

alexanderwiederin:
Absolutely, that was something I suggested in #20. I'll close this and implement `as_bytes -> &[u8]` as a standalone method.

PR #190, opened 2026-05-27
feat(core): Add zero-copy as_bytes to ScriptPubkeyExt

### LINKED ISSUE #20: Feat: Script Type Detection

alexanderwiederin:
closes #20 

## Changes

Adds a zero-copy `as_bytes` method to `ScriptPubkeyExt` that returns a slice directly into kernel-managed memory valid for the lifetime of the script object.

## Motivation

`to_bytes` allocates a `Vec<u8>` on every call. For callers that only need to inspect or pass along the raw bytes without owning them, `as_bytes` avoids that allocation.

## Why this is not applied to other types

Scripts are opaque byte sequences - `as_bytes` is the complete and useful representation. Types like `Block` and `Transaction` have internal structure that callers need to deserialise to use meaningfully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added `Block::check` to perform context-free validation of a block (size, weight, coinbase, transactions, sigops), with optional proof-of-work and merkle-root checks toggled via the `BLOCK_CHECK_BASE` / `_POW` / `_MERKLE` / `_ALL` flags. Returns a `BlockCheckResult` enum carrying the validation state on failure.
+- Added `ScriptPubkeyExt::as_bytes` to return a zero-copy slice into kernel-managed memory. Unlike `to_bytes`, this does not allocate.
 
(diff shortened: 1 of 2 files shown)


rustaceanrob:
LGTM. For the use case of silent payments this would save a vector allocation for each output script that is not a P2TR output, very useful.

sedited (approved):
I think this is the correct way to implement this from the Rust side, but it does bake it into the contract that the kernel library itself does not first serialize the bytes into a temporary. In this case, I think this is fine, and we can probably rely on it for the time being.

LGTM

PR #192, opened 2026-06-05
feat: add custom signet support

oleonardolima:
I've been discussing and working with @edilmedeiros in @vinteumorg study sessions, and as mentioned in https://github.com/bitcoin/bitcoin/issues/35362, currently there's no way to use `rust-bitcoinkernel` with a custom signet.

As implemented in https://github.com/bitcoin/bitcoin/pull/35464, there's now a new method to create the chain parameters with a custom signet challenge through the bitcoinkernel API.

In the latest study session we're reviewing https://github.com/bitcoin/bitcoin/pull/35464 and implemented the last commit below to expose and use the new method through `rust-bitcoinkernel` API.

cc @johnnyasantoss @Davidson-Souza


sedited:
Can you pull in the latest bitcoin core subtree and re-apply your changes now that the upstream change has landed?

sedited (approved):
Thanks, lgtm.

PR #195, opened 2026-06-25
fix(contrib): fix subtree commit hash extraction

alexanderwiederin:
## Changes
Only parse the subject line of commit messages and relax the hash regex to match variable-length git abbreviations.


sedited (approved):
lgtm

PR #196, opened 2026-06-26
feat(core): add `Transaction::check` method for context-free validation

alexanderwiederin:
## Summary

This PR adds the `Transaction::check` method to conduct context-free validation checks on transactions.

### libbitcoinkernel-sys
- Adds the `TxValidationResult` opaque type
- Adds the `btck_TxValidationResult` type alias and named constants
- Adds the `btck_tx_validation_state_` life-cycle methods 
- Adds the `btck_transaction_check` method

### Wrapper
- Adds `Transaction::check` to validate transactions without chain state, returning `TxCheckResult::Valid` or `TxCheckResult::Invalid(TxValidationResult)`

### DIFF
--- a/src/core/transaction.rs
+++ b/src/core/transaction.rs
@@ -149,7 +149,14 @@ use std::{
 
 use libbitcoinkernel_sys::{
     btck_Transaction, btck_TransactionInput, btck_TransactionOutPoint, btck_TransactionOutput,
-    btck_Txid, btck_transaction_copy, btck_transaction_count_inputs,
+    btck_TxValidationResult, btck_TxValidationResult_CONFLICT, btck_TxValidationResult_CONSENSUS,
+    btck_TxValidationResult_INPUTS_NOT_STANDARD, btck_TxValidationResult_MEMPOOL_POLICY,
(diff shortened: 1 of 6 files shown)


src/core/transaction.rs
```diff
@@ -399,6 +408,43 @@ pub trait TransactionExt: AsPtr<btck_Transaction> {
...
+    /// match tx.check()? {
+    ///     TxCheckResult::Valid => println!("transaction passed consensus checks"),
+    ///     TxCheckResult::Invalid(reason) => println!("transaction failed: {:?}", reason),
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn check(&self) -> Result<TxCheckResult, KernelError> {
+        let state = TxValidationState::new();
+        if state.inner.is_null() {
```
sedited: Do we really need to handle a failure here? Afaict this can never fail and our policy with allocation failures is just crash.

alexanderwiederin: Fixed thanks! Also added doc comments for `# Returns`.

sedited (approved):
lgtm

PR #199, opened 2026-06-27
fix(examples): silentpaymentscanner.rs:arg parsing

RandyMcMillan:

PR #200, opened 2026-06-27
chore(ci): examples:cargo run

RandyMcMillan:


RandyMcMillan:
follow up to #199 (combined)

.github/workflows/ci.yml
```diff
@@ -174,8 +174,19 @@ jobs:
         echo "BOOST_INCLUDEDIR=$env:GITHUB_WORKSPACE\vcpkg\installed\x64-windows-static\include" | Out-File -FilePath $env:GITHUB_ENV -Append
         echo "BOOST_LIBRARYDIR=$env:GITHUB_WORKSPACE\vcpkg\installed\x64-windows-static\lib" | Out-File -FilePath $env:GITHUB_ENV -Append
 
-    - name: Run tests
-      run: cargo test -vv
+    - name: Build and Test
+      env:
```
RandyMcMillan, line 178: fixes an intermittent failure:

#### CARGO_HTTP_PROTOCOL: "http1": HTTP/2 is prone to framing layer errors when proxies or unstable networks are involved in CI runners. HTTP/1.1 is significantly more resilient for the small, numerous requests required by cargo.
#### CARGO_HTTP_MULTIPLEXING: "false": This disables the ability to send multiple requests over one connection, which is exactly where the "HTTP2 framing layer" errors usually manifest during concurrent dependency downloads.
#### CARGO_NET_RETRY: Increases the patience of the Cargo client if a transient packet loss occurs during the is_terminal_polyfill download.

.github/workflows/ci.yml
```diff
@@ -109,10 +109,10 @@ jobs:
     - name: Build
       run: cargo build -vv
 
-    - name: Build example binary
+    - name: Run example binary
       run: |
         cd examples
-        cargo build --verbose
+        cargo run --verbose -- datadir
```
alexanderwiederin, line 115: This runs `silentpaymentscanner.rs` against an empty `datadir`, so `scan_chain` will iterate over an empty chain. Does this test anything beyond "the binary doesn't crash on startup"?

.github/workflows/ci.yml
```diff
@@ -174,8 +174,19 @@ jobs:
...
+        CARGO_HTTP_MULTIPLEXING: "false"
+        CARGO_NET_RETRY: "5"
+        CARGO_NET_TIMEOUT: "60"
+      run: |
+        # Clean the stale/corrupted attempt
+        cargo clean
+        # Force an update of the registry without using multiplexing
+        cargo update
+        # Perform the test
+        cargo test -vv
```
alexanderwiederin, line 189: Is this test really a problem? I do not recall seeing a flaky windows run.

alexanderwiederin (commented):
Thanks for the suggestions!

.github/workflows/ci.yml
RandyMcMillan: the flakiness may be due to GitHub infrastructure - not necessarily from the implementation.

.github/workflows/ci.yml
RandyMcMillan: you are correct - please suggest a better implementation.
maybe the simple arg test (above) - then maybe no arg (default) to test default?

RandyMcMillan:
> follow up to #199 (combined)

b3eafcda4cfc82058178a2230808b7f067d4374a updated commit messages per suggestion in #199 

bfc51be1e8bc2d3d54c10a84aecfe6adb6aaab73: remove CHANGELOG.md edit

.github/workflows/ci.yml
alexanderwiederin: I am not sure I understand what you are trying to achieve. I don't think the examples are intended to be run in the CI to check that they run without panicking, but more for people to see what the API looks like in use. What do you think?

PR #201, opened 2026-07-07
feat(script_tracer): add script-tracer feature

alexanderwiederin:
## Summary

Introduces the ScriptTracer behind a `script-tracer` feature flag.

depends on bitcoin/bitcoin/pull/35641


stringintech (commented):
I tried adding trace checks to the existing input verification tests in https://github.com/stringintech/rust-bitcoinkernel/commit/12fe33d2998450b267259c09e8b904e402466a41, mostly to experiment with the feature; feel free to use them if you think it's a good idea!

PR #202, opened 2026-07-08
chore(libbitcoinkernel-sys): note nonnull contract on btck_LogCallback

alexanderwiederin:


sedited:
Why do we need to comment about this in the bindings? Did you run into a potential bug? I do wonder if there would be a better way to express a non-null pointer argument in general in the bindings.

alexanderwiederin:
> Why do we need to comment about this in the bindings? Did you run into a potential bug?

Didn't find a bug, but figured this would be worth pointing out to users of the raw bindings after looking at these more closely implementing the script tracer feature.

> I do wonder if there would be a better way to express a non-null pointer argument in general in the bindings.

I agree. I think we can just remove the `Option<...>` for these non-null arguments.

alexanderwiederin:
Addressed in https://github.com/sedited/rust-bitcoinkernel/pull/204

PR #203, opened 2026-07-08
fix(core): use infallible From for ScriptVerifyStatus

alexanderwiederin:
Status codes are sourced from a vendored subtree build with this crate, so an unknown value indicates a bindings bug, not runtime skew. Panic via `From` rather than returning `KernelError::Internal`.


sedited (approved):
Nice find, lgtm

PR #204, opened 2026-07-09
fix(libbitcoinkernel-sys): make `btck_LogCallback` and `btck_WriteBytes` non-nullable 

alexanderwiederin:
## Summary

Both types are marked `BITCOINKERNEL_ARG_NONNULL` in the C Header at every call site, but were bound as `Option<unsafe extern "C" fn(...)` letting callers pass None, which lowers to a null pointer and is UB in C.

This PR drops the `Option` wrapper so null is representable at the type level.

## Changes

- `btck_LogCallback`: bare `unsafe extern "C" fn(...)`. Updated `btck_logging_connection_create` call site to pass `log_callback::<T>` instead of `Some(log_callback::<T>)`.
- `btck_WriteBytes`: same treatment. Used as the nonnull `writer` arg in `btck_transaction_to_bytes`, `btck_script_pubkey_to_bytes`, `btck_block_to_bytes`. The `c_serialize` helper and its call sites already pass the callback unwrapped, so no further changes needed.

## Why

Neither type ever had a valid "null" use case.


sedited (approved):
ACK 481b055ac050112fcef84e0fc5d2ef16e1c42a77

PR #208, opened 2026-07-17
fix(libbitcoinkernel-sys): add `#[must_use]` parity between `bitcoinkernel.h` and the rust bindings

alexanderwiederin:
## Summary

Add `#[must_use]` parity between `bitcoinkernel.h` and the Rust bindings, enforced in CI.

C's `BITCOINKERNEL_WARN_UNUSED_RESULT` (mostly on `*_create`/`*_copy` functions returning owned pointers) never made it into the Rust bindings.

## Changes

`fix(libbitcoinkernel-sys)`: adds `#[must_use]` to every lib.rs function whose header declaration is `BITCOINKERNEL_WARN_UNUSED_RESULT.`


sedited:
I'm not sure about adding a script for enforcing that. This is just a rough matcher, not something that actually understands the semantics, so it could make us a bit more complacent while something might slip past it. Maybe it would be better to have a checklist of things to verify when updating the header?

alexanderwiederin:
Reduced the scope of this PR to the `must_use` annotation addition. Will think about how to enforce this.

sedited (approved):
LGTM

PR #209, opened 2026-07-21
feat(build): skip building when prebuilt bitcoinkernel is present.

jaoleal:
When LIBBITCOINKERNEL_LIB_DIR is set, skip the entire CMake configure/build/install and link against the pre-built library directly. This follows the same pattern as openssl-sys with OPENSSL_LIB_DIR and helpa with building corner cases.

makes more sense after #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,17 @@ dependencies. Once setup, run:
 cargo b
 ```
 
+### Using a pre-built `libbitcoinkernel`
+
+If you have a pre-built `libbitcoinkernel.a`, you can skip the CMake build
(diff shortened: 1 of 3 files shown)


alexanderwiederin:
One gap vs. the openssl-sys pattern: openssl-sys probes the header to detect the version and gate on compatibility. libbitcoinkernel exposes no version symbol and doesn't guarantee stability, so a prebuilt .a from a different Core version than the vendored headers would link with no diagnostic.

Is skipping the build a hard requirement for you?

jaoleal:
No

PR #210, opened 2026-07-21
docs: add manual Android cross-compilation instructions

jaoleal:
Add a standalone section documenting how to cross-compile for Android without Nix: NDK prerequisites, Rust target installation, environment setup (PATH, Boost discovery), cargo build commands for both the -sys and higher-level crate, linker binary names table, and API level override.

Depends on #158

### DIFF
--- a/ANDROID.md
+++ b/ANDROID.md
@@ -1,0 +1,77 @@
+# Android Cross-Compilation (manual)
+
+This document describes how to cross-compile `rust-bitcoinkernel` for Android
+without Nix. If you can use Nix, prefer the `nix build` outputs described in
+the [README](README.md#android-cross-compilation) — they pin the NDK, Rust
+toolchains, Boost, and cmake for you.
+
+## Prerequisites
+
+Android NDK (r27+ recommended), cmake, and Boost headers installed on the
+host.
+
+Install the Rust target for the architecture you want:
+
+```bash
+rustup target add aarch64-linux-android    # 64-bit ARM
+rustup target add armv7-linux-androideabi  # 32-bit ARM
+rustup target add x86_64-linux-android     # x86_64 emulator
+```
+
+## Environment
+
+The NDK toolchain must be on `PATH` so cmake can find the compilers and
+`llvm-ar`. Boost headers must be discoverable by cmake — either set
+`CMAKE_PREFIX_PATH` or symlink them into the NDK sysroot:
+
+```bash
+export ANDROID_NDK_HOME=/path/to/android-ndk
+
+# Detect host platform
+NDK_HOST="linux-x86_64"   # or "darwin-x86_64" on macOS
+
+# Put NDK clang and llvm-ar on PATH
+export PATH="$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/$NDK_HOST/bin:$PATH"
+
+# Option A: point cmake at your Boost installation
+export CMAKE_PREFIX_PATH=/usr/lib/x86_64-linux-gnu/cmake
+
+# Option B: symlink Boost headers into the NDK sysroot
+ln -sf /usr/include/boost \
+  "$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/$NDK_HOST/sysroot/usr/include/boost"
+```
+
+## Building
+
+Build the `-sys` crate (the static `libbitcoinkernel.a`). The NDK cmake
+toolchain file handles cross-compiler selection, so no extra `CC` or linker
+variables are needed for this step:
+
+```bash
+cargo build -p libbitcoinkernel-sys --target aarch64-linux-android --release
+```
+
+To build the higher-level `bitcoinkernel` crate or run tests, Cargo needs the
+NDK clang as the linker. Set the appropriate `CARGO_TARGET_*_LINKER` variable:
+
+```bash
+export CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER=aarch64-linux-android24-clang
+cargo build --target aarch64-linux-android
+```
+
+The linker binary names for each target are:
+
+| Target                    | Linker                             |
+| ------------------------- | ---------------------------------- |
+| `aarch64-linux-android`   | `aarch64-linux-android24-clang`    |
+| `armv7-linux-androideabi` | `armv7a-linux-androideabi24-clang` |
+| `x86_64-linux-android`    | `x86_64-linux-android24-clang`     |
+
+## API level
+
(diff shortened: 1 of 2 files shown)


jaoleal:
Should i add a entry for this change on CHANGELOG ?

alexanderwiederin:
I think we should have the instructions with nix in `README.md`. For the manual cross-compilation I would add a new file (`ANDROID.md`) and link to it in the `README.md`

> Should i add a entry for this change on CHANGELOG ?

Just added the `no-changelog` label.

jaoleal:
Done @alexanderwiederin 

Github is trolling

jaoleal:
Can we have a release after this one ?

sedited:
> Can we have a release after this one ?

Yes, I'll defer back to Alex here though, and we should wait for the CI to come back again in any case.

ANDROID.md
```diff
@@ -0,0 +1,77 @@
+# Android Cross-Compilation (manual)
+
+This document describes how to cross-compile `rust-bitcoinkernel` for Android
+without Nix. If you can use Nix, prefer the `nix build` outputs described in
+the [README](README.md#android-cross-compilation) — they pin the NDK, Rust
+toolchains, Boost, and cmake for you.
+
+## Prerequisites
+
+Android NDK (r27+ recommended), cmake, and Boost headers installed on the
```
alexanderwiederin, line 10:
```suggestion
Android NDK (tested with r27.3), cmake, and Boost headers installed on the
```

ANDROID.md
```diff
@@ -0,0 +1,77 @@
...
+rustup target add aarch64-linux-android    # 64-bit ARM
+rustup target add armv7-linux-androideabi  # 32-bit ARM
+rustup target add x86_64-linux-android     # x86_64 emulator
+```
+
+## Environment
+
+The NDK toolchain must be on `PATH` so cmake can find the compilers and
+`llvm-ar`. Boost headers must be discoverable by cmake — either set
+`CMAKE_PREFIX_PATH` or symlink them into the NDK sysroot:
```
alexanderwiederin:
```suggestion
`ANDROID_NDK_HOME` selects the NDK that the build uses. If you have more than on NDK installed - Android Studio keeps several - make sure the `PATH` entry below points at the _same_ one, or cmake and the linker will silently use different toolchains.
```

ANDROID.md
```diff
@@ -0,0 +1,77 @@
...
+# Put NDK clang and llvm-ar on PATH
+export PATH="$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/$NDK_HOST/bin:$PATH"
+
+# Option A: point cmake at your Boost installation
+export CMAKE_PREFIX_PATH=/usr/lib/x86_64-linux-gnu/cmake
+
+# Option B: symlink Boost headers into the NDK sysroot
+ln -sf /usr/include/boost \
+  "$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/$NDK_HOST/sysroot/usr/include/boost"
+```
```
alexanderwiederin:
````suggestion
```bash
export ANDROID_NDK_HOME=/path/to/android-ndk

# Host platform tag used in the NDK's prebuilt path.
# Note: NDK prebuilts ship as darwin-x86_64 even on Apple Silicon.
NDK_HOST="linux-x86_64"   # or "darwin-x86_64" on macOS

# Put the NDK toolchain on PATH. Cargo resolves the linker below by bare
# binary name, so it has to be findable on PATH.
export PATH="$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/$NDK_HOST/bin:$PATH"
```

### Boost

Boost headers must be reachable by the NDK compiler. Which of the two options
below you need depends on where Boost is installed.

**Option A — Boost installed outside `/usr`** (a source build, a custom
prefix, or Homebrew on macOS). Point cmake at it:

```bash
export CMAKE_PREFIX_PATH=/opt/boost
```

**Option B — Boost installed by your distro's package manager**, with headers
under `/usr/include/boost`. Symlink them into the NDK sysroot:

```bash
ln -sfn /usr/include/boost \
  "$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/$NDK_HOST/sysroot/usr/include/boost"
```

Option A is **not** sufficient for a distro-installed Boost. cmake finds the
package and configures successfully, but the headers never reach the compiler
and the build fails partway through with:

```
src/txmempool.h:27:10: fatal error: 'boost/multi_index/hashed_index.hpp' file not found
```

cmake omits `/usr/include` from its include flags because it is a standard
system directory. Your host compiler searches it implicitly; the NDK clang
does not, since its sysroot is the NDK's. Use Option B in that case.
````

Comment (alexanderwiederin) [inline ANDROID.md:42]: How about?


## Environment

`ANDROID_NDK_HOME` selects the NDK that the build uses. If you have more than on NDK installed - Android Studio keeps several - make sure the `PATH` entry below points at the _same_ one, or cmake and the linker will silently use different toolchains.

```bash
export ANDROID_NDK_HOME=/path/to/android-ndk
 
NDK_HOST="linux-x86_64"   # or "darwin-x86_64" on macOS
 
export PATH="$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/$NDK_HOST/bin:$PATH"
```

### Boost
 
cmake has to find the Boost package, and the NDK compiler has to find the
headers. For a distro-installed Boost, with headers in `/usr/include/boost`,
that takes two steps:
 
```bash
export CMAKE_PREFIX_PATH=/usr/lib/x86_64-linux-gnu/cmake   # Debian/Ubuntu
 
ln -sfn /usr/include/boost \
  "$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/$NDK_HOST/sysroot/usr/include/boost"
```

Which of the two options (A or B) did you take? Neither original option worked alone for Debian/Ubuntu for me.
 
Without `CMAKE_PREFIX_PATH`, cmake fails at configure. Without the symlink,
configure succeeds and the build fails later with
`'boost/multi_index/hashed_index.hpp' file not found`.
 
If Boost is installed under its own prefix (a source build, or Homebrew on
macOS), pointing `CMAKE_PREFIX_PATH` at that prefix may be enough on its own,
since the headers are then outside the system include directories. If the
build still reports a missing Boost header, add the symlink as well.

---

Which original option (A or B) did you take? Neither original option worked alone for me on Debian/Ubuntu.

alexanderwiederin (commented):
Nice! I would suggest we add a ci job like [this one](https://gist.github.com/alexanderwiederin/28717075775a3d39b4a3a84008b44bf7).

Also suggested something in the environment/boost section - question is inline.

PR #211, opened 2026-07-24
refactor(libbitcoinkernel-sys): restructure `build.rs` around TargetConfig

alexanderwiederin:
## Summary

Replace the inline per-platform conditionals with a TargetConfig struct returned from target_config(), where each target-OS arm supplies its own cmake args, link directives, and rerun-if-env-changed vars. Android handling moves into an Android struct plus an Abi enum that owns the ANDROID_ABI value and the NDK sysroot triple. Wrap the cmake configure, build and install invocations in a Cmake struct, and hoist the fixed flags into BASE_CMAKE_FLAGS.

**Two small follow-ups on top:**

- Android ABI detection now uses `CARGO_CFG_TARGET_ARCH` rather than the previous mechanism.
- Job count falls back to `available_parallelism` when not otherwise specified.

### Why this matters

The refactor aims to separate and clarify build behavior for different targets.

cc: @jaoleal

### DIFF
--- a/libbitcoinkernel-sys/build.rs
+++ b/libbitcoinkernel-sys/build.rs
@@ -1,187 +1,328 @@
 use std::env;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
-fn main() {
-    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
-    let is_android = target_os == "android";
+const BUILD_CONFIG: &str = "RelWithDebInfo";
+const MIN_ANDROID_API: u32 = 24;
+const BITCOIN_SRC_DIR: &str = "bitcoin";
 
-    let ndk = if is_android {
-        Some(
-            env::var("ANDROID_NDK_HOME")
-                .expect("Android target detected but ANDROID_NDK_HOME is not set"),
-        )
+const BASE_CMAKE_FLAGS: &[&str] = &[
+    "-DBUILD_KERNEL_LIB=ON",
+    "-DBUILD_TESTS=OFF",
+    "-DBUILD_BENCH=OFF",
+    "-DBUILD_KERNEL_TEST=OFF",
+    "-DBUILD_TX=OFF",
+    "-DBUILD_WALLET_TOOL=OFF",
+    "-DENABLE_WALLET=OFF",
+    "-DENABLE_EXTERNAL_SIGNER=OFF",
+    "-DBUILD_UTIL=OFF",
+    "-DBUILD_BITCOIN_BIN=OFF",
+    "-DBUILD_DAEMON=OFF",
+    "-DBUILD_UTIL_CHAINSTATE=OFF",
+    "-DBUILD_CLI=OFF",
+    "-DBUILD_FUZZ_BINARY=OFF",
+    "-DBUILD_SHARED_LIBS=OFF",
+    "-DCMAKE_INSTALL_LIBDIR=lib",
+    "-DENABLE_IPC=OFF",
+];
+fn main() {
+    let target = target_config();
+
+    for var in target.rerun_env {
+        println!("cargo:rerun-if-env-changed={var}");
+    }
+
+    let lib_dir = build_kernel(&target.cmake_args);
+
+    println!("cargo:rustc-link-search=native={}", lib_dir.display());
+    println!("cargo:rustc-link-lib=static=bitcoinkernel");
+
+    for directive in &target.link_directives {
+        println!("cargo:{directive}");
+    }
+}
+
+fn build_kernel(extra_cmake_args: &[String]) -> PathBuf {
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let cmake = Cmake::new(BITCOIN_SRC_DIR, &out_dir);
+
+    // Rebuild whenever anything in the Bitcoin Core submodule changes.
+    println!("cargo:rerun-if-changed={BITCOIN_SRC_DIR}");
+
+    cmake.configure(extra_cmake_args);
+    cmake.build();
+    cmake.install();
+
+    cmake.lib_dir()
+}
+
+// Everything that varies by platform, resolved once up front.
+#[derive(Default)]
+struct TargetConfig {
+    // Extra cmake flags needed to cross-compile for this target.
+    cmake_args: Vec<String>,
+    // Emitted verbatim as `cargo:{directive}`.
+    link_directives: Vec<String>,
+    // Variables that should force a rebuild when they change.
+    rerun_env: &'static [&'static str],
+}
+
+fn target_config() -> TargetConfig {
+    match &*env::var("CARGO_CFG_TARGET_OS").unwrap() {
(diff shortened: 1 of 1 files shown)


sedited:
@jaoleal do you want to review this?

libbitcoinkernel-sys/build.rs
```diff
@@ -72,116 +148,186 @@ fn main() {
...
+    println!("cargo:rustc-link-search=native={}", lib_dir.display());
+    println!("cargo:rustc-link-lib=static=bitcoinkernel");
+
+    for directive in target.link_directives() {
+        println!("cargo:{directive}")
+    }
+}
+
+fn build_kernel(target: &dyn Target) -> PathBuf {
+    let bitcoin_dir = Path::new("bitcoin");
```
jaoleal: nit: bitcoind_dir is  constant(maybe it could be externally defined ?) and only used in cmake_configure as source_dir, you could cast it in-place for this commit.

libbitcoinkernel-sys/build.rs
```diff
@@ -1,69 +1,145 @@
...
+        match self {
+            Self::Arm64 => "aarch64-linux-android",
+            Self::ArmV7 => "arm-linux-androideabi",
+            Self::X86_64 => "x86_64-linux-android",
+        }
+    }
+}
+
+struct Android {
+    ndk: PathBuf,
```
jaoleal, line 154: These fields are runtime constants, you don't really need to hold them inside Android as associated getters should return the same value

libbitcoinkernel-sys/build.rs
```diff
@@ -1,69 +1,145 @@
...
+    "-DBUILD_UTIL_CHAINSTATE=OFF",
+    "-DBUILD_CLI=OFF",
+    "-DBUILD_FUZZ_BINARY=OFF",
+    "-DBUILD_SHARED_LIBS=OFF",
+    "-DCMAKE_INSTALL_LIBDIR=lib",
+    "-DENABLE_IPC=OFF",
+];
 
-    println!("{} {}", bitcoin_dir.display(), build_dir.display());
+trait Target {
```
jaoleal: Why this need to be a trait ? Its not like we need to work over given generics here.

As im seeing this can be a finite enum like

```Rust
enum Target {
Windows
MacOs
Android
Unix
}
```

Perhaps this approach would also make it easier to follow and understand the build process ? Am i losing something ?

libbitcoinkernel-sys/build.rs
```diff
@@ -72,116 +148,186 @@ fn main() {
...
-            // CMAKE_PREFIX_PATH. Override it so Boost headers can be located.
-            .arg("-DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=BOTH");
+        out
     }
 
-    cmake_configure
+    fn rerun_env(&self) -> &[&str] {
+        // Without these, changing either variable between builds won't trigger a
+        // rebuild, so we could silently end up with a stale binary built against
+        // the wrong NDK or API level.
```
jaoleal: nit: Now that we have a clearer design for the build process I feel that this comment is pretty much redundant

libbitcoinkernel-sys/build.rs
```diff
@@ -1,69 +1,145 @@
...
+        &[]
+    }
+}
+
+fn target_from_env() -> Box<dyn Target> {
+    match &*env::var("CARGO_CFG_TARGET_OS").unwrap() {
+        "windows" => Box::new(Windows),
+        "macos" => Box::new(MacOs),
+        "android" => Box::new(Android::from_env()),
+        _ => Box::new(Unix),
```
jaoleal: I would change this catch-all to be more explicit, not defaulting to Unix but panicking when we cant recognize a Target that we **explicitly** support

https://doc.rust-lang.org/reference/conditional-compilation.html#target_os says that we can have a bunch of target OS's and this catch-all throws "ios" and "linux" into the same bucket

libbitcoinkernel-sys/build.rs
```diff
@@ -72,116 +148,186 @@ fn main() {
...
+    let install_dir = Path::new(&out_dir).join("install");
 
+    // Iterate through all files in the Bitcoin Core submodule directory
+    println!("cargo:rerun-if-changed={}", bitcoin_dir.display());
+
+    cmake_configure(target, bitcoin_dir, &build_dir, &install_dir);
+    cmake_build(&build_dir);
+    cmake_install(&build_dir);
+
+    resolve_lib_dir(&install_dir)
```
jaoleal: Like what you did with the target os handling, you can do the same thing for the cmake handling, wiring those loose methods as associated ones under a `Cmake` struct

jaoleal (commented):
Nice! really liked what this pr is doing, here some thoughts and suggestions

libbitcoinkernel-sys/build.rs
alexanderwiederin: They are constant for the process agreed, but I think that similarly argues for reading them once. `ndk` for example is used three times. Associated getters would mean three `ANDROID_NDK_HOME` reads each with their own panic path. The idea is to have `from_env()` be a parse step rather than a cache; once it returns nothing downstream has to check them for validity.

libbitcoinkernel-sys/build.rs
alexanderwiederin: Not sure about `enum` either because it scatters the config across the script, but I agree that a `trait` is also not ideal. Will get back to you on this.

libbitcoinkernel-sys/build.rs
alexanderwiederin: Good idea!

alexanderwiederin:
Thanks for the feedback! Will make a push in a few days.

libbitcoinkernel-sys/build.rs
jaoleal: https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html is an option ?

Im pretty much sure we cant use it because of msrv but would be nice to keep an eye on it.

We could work into something like:
```Rust
static NDK: LazyLock<PathBuf> = LazyLock::new(|| { /* ... */ });
static ABI: LazyLock<Abi> = LazyLock::new(Abi::from_env);
static API_LEVEL: LazyLock<u32> = LazyLock::new(|| { /* ... */ });
```

libbitcoinkernel-sys/build.rs
jaoleal: It doesnt need to be perfect, the trait approach is better than the big main() that we had before

libbitcoinkernel-sys/build.rs
alexanderwiederin: `LazyLock` would work, but I don't think it buys much here. The payoff is usually amortizing expensive init across unpredictable call sites, which doesn't really to us here.

jaoleal (approved):
ACK 1220542edde18afc7651679424be953885ce91f2

sedited (approved):
Thanks, some nice improvements in here. LGTM

PR #213, opened 2026-08-05
fix(state): pass the block hash pointer to get_block_tree_entry

### LINKED ISSUE #212: get_block_tree_entry returns None for every hash
Found while debugging a reorg failure downstream on kernel-node. Using the call to look up the active tip's own hash on every received block gave 573,451 lookups, but zero successes.

I noticed that `ChainstateManager::get_block_tree_entry` never returns `Some`, including for hashes of blocks the chainstate manager has just connected.

`BlockHash` is a handle wrapper (`{ inner: *mut btck_BlockHash }`), but the call site casts the address of the wrapper instead of passing the handle it owns: 
```rust
block_hash as *const _ as *const btck_BlockHash
```
I have a fix with regression tests and will open a PR shortly.

pzafonte:
`BlockHash` is a handle wrapper around a `*mut btck_BlockHash`, but the call site cast the address of the wrapper rather than passing the handle it owns. Dereferencing one level short, C read that pointer value where it expected 32 hash bytes, matched nothing in the block index and returned null, which maps to `None`. It fails silently because the address is valid readable memory, and `None` looks the same as an honest miss.

The fix uses the existing accessor, matching every other FFI call in the crate:

```rust
btck_chainstate_manager_get_block_tree_entry_by_hash(self.inner, block_hash.as_ptr())
```
Adds regression tests. One resolves the active chain tip by its own hash and checks that an all-zero hash, an unknown hash, and the tip's hash with one bit flipped all miss. The other walks the chain from tip to genesis and asserts every block resolves to itself, so a lookup returning a fixed entry fails too.

There is no API change as the function returns what it already documented.

Closes #212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `verify` now uses an infallible conversion for the internal `ScriptVerifyStatus`, since an unrecognized status can only indicate a build-time mismatch between the bindings and the vendored `libbitcoinkernel` subtree rather than a runtime condition.
+- `ChainstateManager::get_block_tree_entry` now resolves the requested block instead of returning `None` for every input. It passed the address of the `BlockHash` wrapper to the kernel rather than the block hash handle the wrapper owns, so no lookup ever matched an entry in the block tree.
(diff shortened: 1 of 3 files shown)


sedited (approved):
Thanks, LGTM

PR #214, opened 2026-08-13
chore(core): assert owned FFI wrappers have drop

alexanderwiederin:
A wrapper missing its Drop impl leaks its C handle silently. This adds a test to detect Drop implementation for owned types.


sedited (approved):
LGTM

Removed drop from txid and indeed:
```
---- core::transaction::tests::test_txid_requirements stdout ----

thread 'core::transaction::tests::test_txid_requirements' (185252) panicked at src/core/transaction.rs:1911:5:
owned FFI wrapper must impl Drop to release its handle
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #215, opened 2026-08-14
Fix(core): Fix `TxOutIter::size_hint` to use output count

alexanderwiederin:
### Summary

`TxOutIter::size_hint` computed its remaining count from `input_count()` rather than `output_count()`, so on any transaction where the two differ it reported the wrong number of remaining items.

### Why this wasn't caught

Neither iterator had any coverage of `size_hint` or `len`. The existing `test_transaction_outputs_iterator` drives a manual `for` loop with a hand-rolled counter, which only exercises `next()`, so the hint was invisible to the suite.


sedited (approved):
Uff, nice catch.

LGTM

jaoleal (approved):
LGTM 6e73cd37ab3d19d14422748dbadd83a59f5c9c2c

i can reack on rebase

jaoleal (approved):
ACK aec4a436cdedb9912f8f542f9006daa963459994

sedited (approved):
Will see that I can make a release tomorrow.

PR #219, opened 2026-09-03
chore(flake): bump nixpks 25.06 to 26.05

alexanderwiederin:
25.05 is EOL. 26.05 also carries the static.crates.io switch for importCargoLock, fixing the crates.io 403s in CI - the API host began rejecting curl's default User-Agent, and fetchurl shells out to curl.

cc: @jaoleal

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -41,16 +41,15 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


jaoleal (approved):
ACK cb583a3704a33b168989cbc520c67ba358e8f0e7

PR #221, opened 2026-09-04
build: set `DCMAKE_POSITION_INDEPENDENT_CODE` to `ON`

### LINKED ISSUE #220: Enable Position-independent code build flag
When attempting to cross-compile [`bdk-floresta`](https://github.com/luisschwab/bdk-floresta) to `riscv64gc-unknown-linux-gnu`, the RISC-V linker failed when linking `libbitcoinkernel.a` into `libfloresta_chain.so`:

```
relocation R_RISCV_TPREL_HI20 against `a local symbol' can not be used when making a shared object; recompile with -fPIC
unresolvable R_RISCV_CALL_PLT relocation against symbol `prctl@@GLIBC_2.27'
collect2: error: ld returned 1 exit status
```

Bitcoin Core will enable PIC if the PIE probe assertion is ok, but my cross-compilation CI setup fails in this step. Setting `DCMAKE_POSITION_INDEPENDENT_CODE=ON` fixed it for me, but the fix should be upstreamed here so other downstream consumers don't run into it.

luisschwab:
Closes #220 

Bitcoin Core relies on a PIE probe to decide whether it will enable PIC during build. Some setups may fail the PIE probe (e.g. cross-compilation setups), which in turn will not enable PIC. This lead to failure to link with some rust compilers and rust `crate-type`s, such as `riscv64gc-unknown-linux-gnu` and `cdylib`.

Explicitly setting `DCMAKE_POSITION_INDEPENDENT_CODE=ON` resolves the issue.

### DIFF
--- a/libbitcoinkernel-sys/CHANGELOG.md
+++ b/libbitcoinkernel-sys/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Enabled `CMAKE_POSITION_INDEPENDENT_CODE` when building the kernel. Previously `libbitcoinkernel.a` could not be linked into a shared object on toolchains where Bitcoin Core's PIE probe fails
+
 ## [0.4.0] - 2026-08-26
 
 ### Added
--- a/libbitcoinkernel-sys/build.rs
+++ b/libbitcoinkernel-sys/build.rs
@@ -23,8 +23,10 @@ const BASE_CMAKE_FLAGS: &[&str] = &[
     "-DBUILD_FUZZ_BINARY=OFF",
     "-DBUILD_SHARED_LIBS=OFF",
     "-DCMAKE_INSTALL_LIBDIR=lib",
+    "-DCMAKE_POSITION_INDEPENDENT_CODE=ON",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Enable Position-independent code on the `libbitcoinkernel-sys` build script
```
alexanderwiederin: Rather than in the wrapper's `CHANGELOG`, I think this belongs to `libbitcoinkernel-sys/CHANGELOG` under a `Fixed` section.

```md
### Fixed
- Enabled `CMAKE_POSITION_INDEPENDENT_CODE` when building the kernel. Previously `libbitcoinkernel.a` could not be linked into a shared object on toolchains where Bitcoin Core's PIE probe fails
```

How do you feel about this phrasing?

alexanderwiederin (commented):
Thanks for the fix! 

If you have time, a CI job cross-compiling to `riscv64gc-unknown-linux-gnu` and linking a `cdylib` would be great to prevent regressions, but no pressure.

CHANGELOG.md
luisschwab: Sounds good

PR #224, opened 2026-09-10
feat(workflows): add cross-compilation workflow

alexanderwiederin:
Cross builds are only exercised by the Nix Android job today, so nothing catches breakage that needs a foreign triple. #220 was that: linking a cdylib for riscv64gc failed because Core's PIE probe does not survive a cross setup, leaving PIC off.

Add .github/workflows/cross.yml, building and testing five foreign Linux triples under qemu-user. Three are big-endian, covering byte-order assumptions across the FFI boundary. Each target also links a cdylib; the toolchain file disables the PIE probe, so this reproduces the #220 conditions and will fail if the fix regresses.

Move the Android job out of ci.yml unchanged.

Adapted from https://github.com/luisschwab/bdk-floresta/pull/183

cc: @luisschwab

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -1,0 +1,184 @@
+name: Cross
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+  schedule:
+    - cron: '0 0 * * 1' # weekly on Monday at midnight UTC
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  # Cross-compile the kernel and the Rust bindings for a foreign Linux triple
+  # with a GNU cross toolchain, then run the test suite under qemu-user.
+  #
+  # To add a triple: install `gcc-<prefix>` / `g++-<prefix>` from the Ubuntu
+  # archive, and add a matrix entry. `sysroot` is always /usr/<prefix>.
+  #
+  # The cross setup below - host Boost headers, the cmake toolchain file, and
+  # the qemu-user runners - is adapted from bdk-floresta's cross workflow:
+  # https://github.com/luisschwab/bdk-floresta/pull/183
+  linux:
+    name: ${{ matrix.target }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - target: riscv64gc-unknown-linux-gnu
+            prefix: riscv64-linux-gnu
+            processor: riscv64
+            emulator: qemu-riscv64-static
+
+          - target: powerpc64le-unknown-linux-gnu
+            prefix: powerpc64le-linux-gnu
+            processor: ppc64le
+            emulator: qemu-ppc64le-static
+
+          # Big-endian. These exercise byte-order assumptions across the FFI
+          # boundary that no little-endian target can catch.
+          - target: s390x-unknown-linux-gnu
+            prefix: s390x-linux-gnu
+            processor: s390x
+            emulator: qemu-s390x-static
+
+          - target: powerpc64-unknown-linux-gnu
+            prefix: powerpc64-linux-gnu
+            processor: ppc64
+            emulator: qemu-ppc64-static
+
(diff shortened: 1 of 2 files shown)


.github/workflows/cross.yml
```diff
@@ -0,0 +1,179 @@
...
+            processor: ppc64
+            emulator: qemu-ppc64-static
+ 
+          - target: sparc64-unknown-linux-gnu
+            prefix: sparc64-linux-gnu
+            processor: sparc64
+            emulator: qemu-sparc64-static
+ 
+    steps:
+      - uses: actions/checkout@v4
```
luisschwab, line 60: Any reason to not use a locked hash? If you add a zizmor workflow (eg https://github.com/luisschwab/bdk-floresta/blob/master/.github/workflows/audit.yml#L76), it will complain about this.

.github/workflows/cross.yml
```diff
@@ -0,0 +1,179 @@
...
+ 
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
```
luisschwab: GH runners already have rust installed, so this is unnecessary. You could use a vendored action that sets up `cargo-rbmt` (eg https://github.com/luisschwab/bdk-floresta/blob/master/.github/actions/setup-rbmt/action.yml) and run it on all jobs that require a rust toolchain, as it will automatically install stable, nightly and MSRV.

.github/workflows/cross.yml
```diff
@@ -0,0 +1,179 @@
...
+            echo "CARGO_TARGET_${cargo_suffix}_LINKER=${{ matrix.prefix }}-gcc"
+            echo "CARGO_TARGET_${cargo_suffix}_RUNNER=${{ matrix.emulator }} -L $sysroot"
+          } >> "$GITHUB_ENV"
+ 
+      - name: Build
+        run: cargo build -p bitcoinkernel --target ${{ matrix.target }}
+ 
+      - name: Link as cdylib
+        run: |
+          cargo rustc -p bitcoinkernel --target ${{ matrix.target }} --crate-type cdylib
```
luisschwab: The crate types can be set directly on `Cargo.toml`: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#library

.github/workflows/cross.yml
```diff
@@ -0,0 +1,179 @@
...
+          # required by bionic on x86_64. Tests are skipped until the
+          # MSRV is bumped or a workaround is found. See #198
+          - libbitcoinkernel-android-x86_64
+    steps:
+    - uses: actions/checkout@v4
+ 
+    - name: Install Nix
+      uses: cachix/install-nix-action@v31
+ 
+    - name: Build${{ matrix.package != 'libbitcoinkernel-android-x86_64' && ' and Test' || '' }}
```
luisschwab:
```suggestion
    - name: Build ${{ matrix.package != 'libbitcoinkernel-android-x86_64' && ' and Test' || '' }}
```

luisschwab (commented):
A few comments

.github/workflows/cross.yml
alexanderwiederin: Dropped the cargo rustc approach in the force-push. The crate exports no C symbols, so a cdylib link barely touches the archive and would have passed either way. It now links the archive directly with --whole-archive, which forces every member in.

.github/workflows/cross.yml
alexanderwiederin: No good reason, just followed the existing. Opened https://github.com/sedited/rust-bitcoinkernel/issues/225 to fix in other places too. Thanks!

.github/workflows/cross.yml
alexanderwiederin: Swapped it for a plain `rustup target add ${{ matrix.target }}`. 

One clarification on the reasoning though: that step wasn't installing Rust, it was there for the `targets:` input.

luisschwab (approved):
ACK cf80f12bedccbf86141fdab3ebf2d9137b956bc3

PR #226, opened 2026-09-14
feat(workflows): drop credential persistence, narrow permissions and add zizmor

### LINKED ISSUE #225: ci: pin actions to commit hashes, add zizmor and Dependabot
see https://github.com/sedited/rust-bitcoinkernel/pull/224#discussion_r3981121196 and

alexanderwiederin:
Closes #225. Follows up on @luisschwab's review comment in #224.

Reduced from the original scope. Two commits:

- Set `persist-credentials: false` on checkout and add `permissions: contents: read` to the workflows
- Add a `zizmor` workflow so those two don't silently regress


.github/workflows/ci.yml
```diff
@@ -18,10 +21,12 @@ jobs:
...
         dep: [minimal, recent]
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+      with:
+        persist-credentials: false
 
     - name: Install Rust MSRV
-      uses: dtolnay/rust-toolchain@master
+      uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # v1
```
luisschwab: If you add the major, minor and patch versions to the comment, dependabot will automatically modify it as well

alexanderwiederin: Thanks! This one can't be Dependabot-tracked at all, so I'll open an issue to decide between dropping the action for a plain `rustup toolchain install`. https://github.com/sedited/rust-bitcoinkernel/issues/227

luisschwab: Why don't you use cargo-rbmt on CI as well? Then you can have a vendored action that sets it up and installs the pinned toolchains and tools. See bdk-floresta, rustreexo or rust-bitcoin repos for examples.

